PR #8, opened 2017-05-02
Add support for a Loading Button

phndiaye:


addon/components/loading-button.js
```diff
@@ -0,0 +1,41 @@
...
+      this.get('targetObject').send(
+        this.get('slowAction'),
+        this.get('params'),
+        defer
+      );
+    }, 100);
+  }
+});
+
+LoadingButtonComponent.reopenClass({
```
AlexisMontagne: why do you reopen the class? you can't no put it in the initial instantiation?

addon/components/loading-button.js
```diff
@@ -0,0 +1,41 @@
...
+
+  click(e) {
+    e.preventDefault();
+
+    // Because `.send` method for sending actions does not return anything, we
+    // pass it an `RSVP.defer` to be resolved in the remote action.
+    let defer = RSVP.defer();
+    defer.promise.then(() => {
+      this.$().text(this.get('originalContent'));
+    }, () => {
```
AlexisMontagne: can you  skip this empty param?

phndiaye: Actually i wanted to ask if you have anything in mind if there are error ... is it to the button to handle it ? 

(And that part is skippable, yes)

AlexisMontagne: i think the error will be handled by the caller (at least at the beginning)

addon/components/loading-button.js
```diff
@@ -0,0 +1,41 @@
...
+
+    });
+
+    Ember.run.debounce(this, function() {
+      this.$().width(this.$().width());
```
tgallice: This is for what?

addon/components/loading-button.js
phndiaye: For unknown reasons ... that's how Ember documents this (weird af, i know)

phndiaye: (Tested the param in the initial instantiation)

addon/components/loading-button.js
```diff
@@ -0,0 +1,41 @@
...
+      this.set('originalContent', this.$().html());
```
tgallice: `html` ? `text` will be more consistent with L18

addon/components/loading-button.js
phndiaye: The button's width adapts to its content, which gives a weird effect the content is just the spinner. This just keeps the initial size while loading

tgallice: This implementation seems really wierd. To me extract the `width` and `originalContent` in variables (they can be keep in the function scope) at the top of the function willl be more readable.

tests/dummy/app/templates/partials/buttons/stateful.hbs
```diff
@@ -0,0 +1,23 @@
...
+
+  {{#destructive-button destructiveAction='deleteStuffWithError' record='hey'
+                        ongoingMessage='Cancel'
+                        successMessage='Deleted!'
+                        failureMessage='Failed!'
+                        class='margin-left-xx-sm'}}
+    {{fa-icon 'trash'}} &nbsp; Failing Button
+  {{/destructive-button}}
+
+  {{#loading-button 'hello' 'hey' slowAction='runSlowAction'
```
tgallice: missing `record` no ?

phndiaye: Actually i wanted to ask you about it. The problem with "named" parameters is that you need to know them all. Until now, passing `record` only works, but what if we want to do something that does not take a record but something else (or more than one parameters) ?

With the positional params, we can pass as many parameters to the action we want as it will be an array (to be deconstructed by the calling application)

addon/components/loading-button.js
```diff
@@ -0,0 +1,46 @@
...
+    e.preventDefault();
+
+    let originalContent = this.$().html();
+    let originalWidth = this.$().width();
+
+    // Because `.send` method for sending actions does not return anything, we
+    // pass it an `RSVP.defer` to be resolved in the remote action.
+    let defer = RSVP.defer();
+    defer.promise.then(() => {
+      this.$().removeAttr('style');
```
tgallice: Why you need to remove `style` ?

phndiaye: Because when i set the `loading button width` to the original button width, it's set a `style="width: ..."`.

tgallice: Yes but since the width is the same this should not be a problem no ?

phndiaye: Yes. But i prefer going back to a "clean" state. No ?

phndiaye:
@AlexisMontagne @tgallice Any additional comment on this one, guys ?

addon/components/loading-button.js
```diff
@@ -0,0 +1,46 @@
+import Ember from 'ember';
+
+const {
+  Component,
+  RSVP,
+  TargetActionSupport
```
AlexisMontagne: What about this one?

phndiaye: Fixed ... but don't know how imhotep reads the file -_-

phndiaye: https://github.com/upfluence/oss-components/pull/8/files#diff-e7fa075c5fbc52f47a6beb477e93137eR9

PR #10, opened 2017-05-03
Use TargetActionSupport

### LINKED ISSUE #9: Destructive button
I think there is a BC break with newness version of ember about the `targetObject` property. Maybe use https://github.com/emberjs/ember.js/blob/v2.12.0/packages/ember-runtime/lib/mixins/target_action_support.js could be a better choise to handle/send the action.

tgallice:
close #9


phndiaye:
Didn't even know this existed ... :o 

LGTM 👍

PR #15, opened 2017-07-25
Add support for running a callback when user click on a row

phndiaye:


addon/components/upf-table/index.js
```diff
@@ -19,6 +19,7 @@ export default Component.extend({
     return this.get('columns').map((column) => {
       column.visible = (column.visible !== false) ? true : column.visible;
       column.sorted = (column.sorted !== true) ? false : column.sorted;
+      column.editable = (column.editable !== false) ? true : column.editable;
```
phndiaye: Just the condition would do it ...

addon/components/upf-table/row.js
```diff
@@ -1,9 +1,38 @@
...
 export default Component.extend({
   tagName: 'tr',
   classNames: ['upf-datatable__row'],
-  classNameBindings: ['isHeaderRow:upf-datatable__row--header']
+  classNameBindings: [
+    'isHeaderRow:upf-datatable__row--header',
+    'isSelected:upf-datatable__row--selected'
+  ],
+
+  isSelected: computed('ref.selected', function() {
```
tgallice: maybe `computed.bool('ref.selected')`

PR #16, opened 2017-07-31
Prefere triggerAction instead of send to the targetObject/_targetObject

tgallice:


tgallice:
This should fix the toggle `paid?` issue

phndiaye:
LGTM 👍

PR #19, opened 2017-08-03
Toggle the pagination

tgallice:


addon/components/upf-table/index.js
```diff
@@ -15,6 +15,7 @@ export default Component.extend({
   allRowsSelected: false,
   displayedColumnsPanel: false,
 
+  isPaginated: false,
```
phndiaye: would have put it just after the `hasSortableColumns` so it stays with the table options.
+ had called it `hasPagination`, so that we keep the `hasX` naming :)

phndiaye:
LGTM 👍

PR #20, opened 2017-08-03
Datatable search support

phndiaye:


app/templates/components/input-wrapper.hbs
```diff
@@ -1,19 +1,17 @@
...
-    <span class='col-xs-12 upf-input-feedback upf-input-feedback--error'>
-      <i class='fa fa-exclamation-circle'
-         aria-label={{error}}>
+{{#if error}}
+  <span class='col-xs-12 upf-input-feedback upf-input-feedback--error'>
+    <i class='fa fa-exclamation-circle'
+       aria-label={{error}}>
+    </i>
+  </span>
+{{else}}
```
tgallice: useless `else` IMO

app/templates/components/input-wrapper.hbs
```diff
@@ -1,19 +1,17 @@
...
+  {{#if help}}
```
tgallice: can be `else if`

PR #21, opened 2017-08-03
Add basic loading state to the datatable

phndiaye:


addon/components/upf-table/index.js
```diff
@@ -39,6 +40,11 @@ export default Component.extend({
     });
   }),
 
+
+  _initiallyDisplayedColumns: computed.filter('_columns', function(column, _, _) {
```
tgallice: Maybe just `computed.filterBy('_columns', 'visible')`

PR #28, opened 2017-10-11
Fix disabled

tgallice:
Currently the `disabled` state is never removed after the resolving of the promise due to an none update of the  `shouldBeDisabled` proprety. I'm not really sure on what going wrong, but I have use an observer to set the `disabled` which will make the thing work.

@AlexisMontagne @phndiaye If you have some insight on it.


AlexisMontagne:
LGTM

phndiaye:
LGTM 👍

PR #30, opened 2017-11-27
Factorize table search into a mixin

phndiaye:


addon/mixins/upf-table-search.js
```diff
@@ -0,0 +1,41 @@
...
+  filteredCollection: computed('searchQuery', function() {
+    let collection = get(this, this.get('searchCollection'));
+
+    if (this.get('searchQuery') === '') {
+      return collection;
+    }
+
+    return collection.filter((item) => {
+      let i;
+      let searchWords = this.get('searchQuery').split(' ');
```
tgallice: this can be extract from the filter method

addon/mixins/upf-table-search.js
```diff
@@ -0,0 +1,41 @@
...
+
+      for (i = 0; i < searchWords.length; i++) {
+        let searchMatch = item.get(
+          this.get('searchAttribute')
+        ).toLowerCase().includes(searchWords[i].toLowerCase());
```
tgallice: Same, can be extracted

addon/mixins/upf-table-search.js
```diff
@@ -0,0 +1,41 @@
...
+      return collection;
+    }
+
+    return collection.filter((item) => {
+      let i;
+      let searchWords = this.get('searchQuery').split(' ');
+
+      for (i = 0; i < searchWords.length; i++) {
+        let searchMatch = item.get(
+          this.get('searchAttribute')
```
tgallice: this can be move before the loop

PR #35, opened 2018-01-03
Upf Stat component

phndiaye:


addon/components/upf-stat.js
```diff
@@ -0,0 +1,16 @@
...
+
+const { Component } = Ember;
+
+export default Component.extend({
+  classNames: ['upf-stat'],
+
+  name: null,
+  data: null,
+
+  progressValue: null,
```
tgallice: This data should be calculated before. It's not better to have a maxValue and compute the progress in the component ?

ramshorst:
Great you undertook this!

Some feedback : The centering is not equal some times :
screen shot 2018-01-04 at 09 56 18

bower.json
```diff
@@ -5,7 +5,7 @@
     "ember-cli-shims": "0.1.1",
     "ember-cli-test-loader": "0.2.2",
     "ember-qunit-notifications": "0.1.0",
-    "upfluence-oss": "0.4.0",
+    "upfluence-oss": "git://github.com/upfluence/oss.git#pn-upf-stat-and-progress",
```
tgallice: should be revert no ?

PR #37, opened 2018-01-16
Enhancing upf-stats

ramshorst:


addon/components/upf-stat.js
```diff
@@ -1,13 +1,28 @@
...
   data: null,
+  dataClass: null,
+  label: null,
 
   icon: null,
-  iconClass: ''
+  iconClass: '',
+
+  dataClass: computed('data', function() {
+    if (this.get('data') !== 0) {
```
phndiaye: `if (this.get('data') === 0)` and no need to handle the `else` case

addon/components/upf-stat.js
```diff
@@ -1,13 +1,28 @@
...
+      return null
+    } else {
+      return ' upf-stat__data__null'
```
phndiaye: `upf-stat__data--null`

phndiaye:
LGTM 👍

phndiaye:
LGTM

PR #38, opened 2018-01-22
Enable link and Tooltip on icon

ramshorst:


addon/components/upf-stat.js
```diff
@@ -15,5 +15,11 @@ export default Component.extend({
   label: null,
 
   icon: null,
-  iconClass: ''
+  iconClass: '',
+  iconUrl: null,
+  iconLabel: null,
+
+  didRender() {
+    this.$('[title]').tooltip({ placement: 'bottom' });
```
AlexisMontagne: may be wrap it in: `if (!isBlank(this.get('iconLabel')) {`

ramshorst: Great idea :) thanks

phndiaye:
LGTMish 😞

PR #49, opened 2019-01-16
Use OSS as an NPM dependency

phndiaye:
- [x] Handle Bootstrap and OSS inclusion into the Ember projects

Depends on: https://github.com/upfluence/oss/pull/58


index.js
```diff
@@ -69,7 +80,7 @@ module.exports = {
     publicAssets.forEach((assetType) => {
       trees.push(
         new Funnel(
-          `${this.app.bowerDirectory}/upfluence-oss/${assetType}/`,
+          `node_modules/upfluence-oss-components/node_modules/upfluence-oss/${assetType}/`,
```
tgallice: in the final project, upfluence-oss will be not installed at the root of the project node_module dir ?

phndiaye: nope, the only way to do that is to have `upfluence-oss` as a direct dependency of the final project :/

tgallice: hmm it's wierd because the most of packages in the node_module are a direct dependency of the final projet

package.json
```diff
@@ -29,7 +29,8 @@
     "ember-cli-shims": "^1.2.0",
     "ember-cli-toggle": "^3.0.0",
     "ember-font-awesome": "2.1.1",
-    "ember-truth-helpers": "^1.3.0"
+    "ember-truth-helpers": "^1.3.0",
+    "upfluence-oss": "upfluence/oss#pn-bower-deprecation"
```
AlexisMontagne: you can put it upstream now?

phndiaye: yep !

PR #53, opened 2019-06-13
Allow disabled option for radio button segmented control

leanminmachine:
While loading if disabled is true, then don't show the isChecked state as well till it is not disabled anymore. 

Wanting to use this for analytics-web because it takes a while for the following count on edit form to update to the correct value. So while it is updating, then i disable the radio buttons, until it is done updating then i enable it and show the correct selection type `currentType`.

### DIFF
--- a/addon/components/radio-button.js
+++ b/addon/components/radio-button.js
@@ -11,7 +11,7 @@ export default Component.extend({
   }),
 
   isChecked: computed('value', 'currentValue', 'disabled', function() {
-      return !this.disabled && this.get('value') === this.get('currentValue');
+      return !this.disabled && this.value === this.currentValue;
   }),
 
   click() {
--- a/addon/components/radio-button.js
+++ b/addon/components/radio-button.js
@@ -11,10 +11,7 @@ export default Component.extend({
   }),
 
   isChecked: computed('value', 'currentValue', 'disabled', function() {
-    if (!this.disabled) {
-      return this.get('value') === this.get('currentValue');
-    }
-    return false;
+      return !this.disabled && this.get('value') === this.get('currentValue');
   }),
 
   click() {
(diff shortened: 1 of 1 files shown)


addon/components/radio-button.js
```diff
@@ -1,19 +1,25 @@
 import Component from '@ember/component';
 import { computed } from '@ember/object';
+import { equal } from '@ember/object/computed';
 
 export default Component.extend({
   tagName: 'label',
   classNames: ['btn', 'upf-radio-btn'],
-  classNameBindings: ['isChecked:active'],
+  classNameBindings: ['isChecked:active', 'isDisabled:disabled'],
```
phndiaye: you can just do `disabled:disabled` and no need to define a `isDisabled` computed.

addon/components/radio-button.js
```diff
@@ -4,14 +4,17 @@ import { computed } from '@ember/object';
...
   attributeBindings: ['style'],
   style: computed('options', function() {
     return `width: ${100 / Object.keys(this.get('options')).length}%;`;
   }),
 
-  isChecked: computed('value', 'currentValue', function() {
-    return this.get('value') === this.get('currentValue');
+  isChecked: computed('value', 'currentValue', 'disabled', function() {
+    if (!this.disabled) {
+      return this.get('value') === this.get('currentValue');
```
AlexisMontagne: `return !this.disable && this.value === this.currentValue` would do it

addon/components/radio-button.js
```diff
@@ -4,14 +4,17 @@ import { computed } from '@ember/object';
 export default Component.extend({
   tagName: 'label',
   classNames: ['btn', 'upf-radio-btn'],
-  classNameBindings: ['isChecked:active'],
+  classNameBindings: ['isChecked:active', 'disabled'],
```
phndiaye: more like `disabled:disabled`, the format is `computedToCheck:attributeToCheck`

phndiaye:
LGTM 👍

PR #68, opened 2020-01-15
Improve slider handling and updates

Hegazy360:


addon/components/upf-slider.js
```diff
@@ -1,14 +1,15 @@
...
+import { notEmpty, or } from '@ember/object/computed';
 
 export default Component.extend({
   classNames: ['upf-slider'],
-  classNameBindings: ['color'],
+  classNameBindings: ['color', 'active'],
 
-  value: 0,
-  active: false,
+  value: or('options.value', '0'),
```
phndiaye: i think `value` can be passed directly given that it's not really an option

PR #71, opened 2020-01-29
Add support for numeric-range component

phndiaye:


addon/components/upf-numeric-range.js
```diff
@@ -0,0 +1,27 @@
...
+
+export default Component.extend({
+  classNames: ['upf-numeric-range'],
+  classNameBindings: ['sizeSmall:upf-numeric-range--small'],
+
+  size: null,
+
+  sizeSmall: equal('size', 'small'),
+
+  defaultOptions: {
```
tgallice: usefull ?

PR #73, opened 2020-05-26
Star Rating Component

Hegazy360:


addon/components/upf-rating.js
```diff
@@ -0,0 +1,11 @@
+import Component from '@ember/component';
+
+export default Component.extend({
+  classNames: ['upf-rating'],
+
+  actions: {
+    saveRatingValue(rating) {
+      this.influencer.set('rating', rating);
```
phndiaye, line 8: OSS is not supposed to be aware of the `influencer`, you should only pass it an object and that's it

phndiaye: and where is that action triggered ?

Hegazy360: Oh yes, that's no longer used, forgot to remove it
Now I pass the callback function as a parameter

PR #92, opened 2021-03-31
Fix Ember build leading to Storybook not loading correctly

phndiaye:
### What does this PR do?

Fix Storybook not building correctly because of the build not working when ember-modifiers is declared as a `dependency` instead of `devDependency`

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled


addon/modifiers/on-click-outside.js
```diff
@@ -1,20 +1,30 @@
...
+        if (!element.contains(event.target)) {
+          callback(state.element, event);
+        }
+      }
 
-  return () => {
-    document.removeEventListener('click', handleClick, false);
-  };
-});
+      console.log('hello')
```
Hegazy360, line 26: image

Hegazy360 (approved):

JulienVannier66 (approved):

PR #99, opened 2021-07-05
Add support for modal component

phndiaye:
### What does this PR do?

Add a modal component here so we can remove the one in upf-utils at some point.

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled


addon/components/o-s-s/modal.js
```diff
@@ -0,0 +1,43 @@
...
+    if (this.args.options?.container) {
+      modal.appendTo(this.args.options.container);
+    }
+
+    this.element.addEventListener('keydown', this._handleEscapeKey.bind(this));
+  }
+
+  @action
+  teardown() {
+    jQuery(this.element).modal('hide');
```
Hegazy360, line 29: why not $?

phndiaye: just imported it under `jQuery`, but yes, possible

Hegazy360 (approved):

addon/components/o-s-s/modal.js
Miexil: 💵

Miexil (approved):

PR #106, opened 2021-09-01
Infinite Select component support

phndiaye:
### What does this PR do?

Adds support for an selection component which supports infinite loading natively. Available features are:

* Search (enabled by default)
* Infinite Loading hook (everytime the bottom of the options is reached the hook is called)
* Simple option display by passing an `itemLabel` (default: "name"). On each item, we will display the value of `item[itemLabel]`
* Complex option by calling the component with a named block as follows:
```
    <OSS::InfiniteSelect @items={{this.items}} @onSelect={{this.onSelect}}>
      <:option as |opt|>
        My custom layout with — {{opt.superAttribute}}
      </:option>
    </OSS::InfiniteSelect>
```

Related to: #

Screenshot 2021-09-01 at 15 35 33

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled


tests/dummy/app/controllers/application.ts
```diff
@@ -0,0 +1,32 @@
+import Controller from '@ember/controller';
+
+export default class Application extends Controller {
+  items = [
```
Miexil, line 4: We're not really using this, but does this work ?
```
items: {name: string}[] =
```

Miexil:
🍾 🥳 🎊  for Typescript ! ❤️

tests/dummy/app/controllers/application.ts
phndiaye: this will disappear before the pr is closed. it's only dummy data to display

addon/components/o-s-s/infinite-select.hbs
```diff
@@ -0,0 +1,21 @@
+<div class="upf-infinte-select upf-floating-menu">
+  {{#if this.searchEnabled}}
+    {{#input-wrapper}}
+      <Input
+        value={{this._searchKeyword}} {{on "keyup" this.updateSearchKeyword}} type="text"
+        class="form-control upf-input" placeholder="Search..." />
```
Miexil, line 6: Shouldn't we put the placeholder either :
1) as a parameter, or 
2) in the en-us.yml ?

tests/dummy/app/controllers/application.ts
Miexil: Yeah I saw it was in the dummy, was just curious actually ^^

Miexil (approved):

tests/integration/components/o-s-s/infinite-select-test.ts
```diff
@@ -0,0 +1,22 @@
...
+    //// Handle any actions with this.set('myAction', function(val) { ... });
+    //await render(hbs`{{o-s-s/infinite-select}}`);
+    //assert.equal(this.element.textContent.trim(), '');
+    //// Template block usage:
+    //await render(hbs`
+    //{{#o-s-s/infinite-select}}
+    //template block text
+    //{{/o-s-s/infinite-select}}
+    //`);
+    //assert.equal(this.element.textContent.trim(), 'template block text');
```
tgallice, line 20: :see_no_evil:

JulienVannier66 (approved):
LGTM

PR #118, opened 2021-09-27
Add toast service

phndiaye:
### What does this PR do?

As part of the war against small libraries, this adds support for a toast service allowing us to trigger toasts in the UI.
For backward compatibility, the service implements the same interface as the one from `ember-toastr`:
- success
- info
- warning
- error

The options have the same names but I just brought back the one we were using being it doesn't need to be a highly configurable library.

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled


addon/services/toast.ts
```diff
@@ -0,0 +1,138 @@
...
+  closeButton: true,
+  onclick: null,
+  tapToDismiss: true
+});
+
+function delegate(ctx: Toast, type: ToastType): Function {
+  return (message: string, title: string, opts: ToastOptions = DEFAULT_OPTIONS): void => {
+    if (isEmpty(message) && isEmpty(title)) return;
+
+    console.log(opts);
```
JulienVannier66, line 30: 🧐

addon/services/toast.ts
```diff
@@ -0,0 +1,138 @@
...
+  private _toasts: Element[] = [];
+
+  success = delegate(this, ToastType.SUCCESS);
+  error = delegate(this, ToastType.ERROR);
+  warning = delegate(this, ToastType.WARNING);
+  info = delegate(this, ToastType.INFO);
+
+  buildToast(type: ToastType, message: string, title: string, opts: ToastOptions): void {
+    const container = this._buildContainer();
+    console.log(opts);
```
JulienVannier66, line 45: 🧐

addon/services/toast.ts
```diff
@@ -0,0 +1,138 @@
...
+      500
+    );
+
+    this._toasts.push(toast);
+    later(this, () => this._destroyToast(toast), TIMEOUT);
+  }
+
+  private _destroyToast(toast: Element | HTMLElement | null): void {
+    if (toast) {
+      toast.removeEventListener('click', () => this._destroyToast(toast), false);
```
JulienVannier66, line 97: It is working to remove event without given a function name (just a arrow function)?

tests/dummy/app/styles/app.less
```diff
@@ -1 +1,61 @@
...
+    margin-right: 50%;
+    transform: translateX(-50%);
+    align-items: center;
+  }
+
+  .toast--hidden {
+    display: flex;
+    opacity: 0;
+    animation: none;
+  }
```
JulienVannier66, line 41: Is it better to use `visibility: hidden`?

addon/services/toast.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    later(
+      this,
+      () => {
+        toast.classList.remove('toast--hidden');
+        toast.classList.add('toast--visible');
+      },
+      500
+    );
```
tgallice, line 89: just to understand, why this ?

addon/services/toast.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  private _destroyToast(toast: Element | HTMLElement | null): void {
+    if (toast) {
+      toast.removeEventListener('click', () => this._destroyToast(toast), false);
+      toast.remove();
+
+      let container = document.querySelector('body #toast-container.upf-toastr--container');
+      if (container?.children.length === 0) {
+        container.remove();
+      }
```
tgallice, line 103: we really need to remove the container each time ?

addon/services/toast.ts
phndiaye: because we can't animate an element being added to the dom (nor can we add `transition` on the CSS `display` rule). So I add it to the dom first, then animate the visibility.

addon/services/toast.ts
phndiaye: it's removed only if there are no toast shown anymore. It was just for cleaning up instead of having an empty div in the dom everytime.

addon/services/toast.ts
tgallice: Nice, didn't see the animation :+1:

addon/services/toast.ts
tgallice: But since we generally display one toast at the time, this will be removed "each" time, no ?

phndiaye: we have some cases where it's not the case, like the import in analytics

tests/dummy/app/styles/app.less
Miexil: FYI, this is probably linked to the animation state. You can't have a proper show/hide CSS animation on the "visibility" attribute whereas it works nicely over "opacity".

addon/services/toast.ts
```diff
@@ -0,0 +1,138 @@
...
+  onclick?: Function | null;
+};
+
+const DEFAULT_OPTIONS: ToastOptions = Object.freeze({
+  closeButton: true,
+  onclick: null,
+  tapToDismiss: true
+});
+
+function delegate(ctx: Toast, type: ToastType): Function {
```
Miexil, line 26: I think the type should actually be 
`function delegate(ctx: Toast, type: ToastType): (message: string, title: string, opts: ToastOptions = DEFAULT_OPTIONS) => void`

Miexil: we might be able to create a type from it. This would allow us to use the type inside your Toast class bellow (for success, error, warning, info "attributes")
I'd like to look into this with you 😇

addon/services/toast.ts
```diff
@@ -0,0 +1,138 @@
...
+  };
+}
+
+export default class Toast extends Service {
+  private _toasts: Element[] = [];
+
+  success = delegate(this, ToastType.SUCCESS);
+  error = delegate(this, ToastType.ERROR);
+  warning = delegate(this, ToastType.WARNING);
+  info = delegate(this, ToastType.INFO);
```
Miexil, line 41: type these with what I was suggesting on the `function delegate` comment ?

addon/services/toast.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  buildToast(type: ToastType, message: string, title: string, opts: ToastOptions): void {
+    const container = this._buildContainer();
```
Miexil, line 44: type :)

addon/services/toast.ts
```diff
@@ -0,0 +1,138 @@
...
+    console.log(opts);
+
+    const toast = document.createElement('div');
+    toast.classList.add('toast', 'toast--hidden', `toast-${type}`);
+
+    const titleContainer = document.createElement('div');
+    titleContainer.classList.add('toast-title');
+    titleContainer.textContent = title;
+
+    const messageContainer = document.createElement('div');
```
Miexil, line 54: nit
but type as well :)

addon/services/toast.ts
```diff
@@ -0,0 +1,138 @@
...
+    this._toasts.push(toast);
+    later(this, () => this._destroyToast(toast), TIMEOUT);
+  }
+
+  private _destroyToast(toast: Element | HTMLElement | null): void {
+    if (toast) {
+      toast.removeEventListener('click', () => this._destroyToast(toast), false);
+      toast.remove();
+
+      let container = document.querySelector('body #toast-container.upf-toastr--container');
```
Miexil, line 100: `container:Element = ` ?

addon/services/toast.ts
```diff
@@ -0,0 +1,138 @@
...
+
+      closeButton.classList.add('toast-close-button');
+      closeButton.innerHTML = '<i class="fas fa-times text-size-5"></i>';
+      closeButton.addEventListener('click', this._onToastClose.bind(this), false);
+
+      toastChildren.push(closeButton);
+    }
+
+    if (opts.tapToDismiss) {
+      toast.addEventListener('click', () => this._destroyToast(toast), false);
```
Miexil, line 71: Might want to save this in order to remove the event listener ?

Miexil: as @JulienVannier66 suggested below 🙂

addon/services/toast.ts
```diff
@@ -0,0 +1,138 @@
...
+      toast.remove();
+
+      let container = document.querySelector('body #toast-container.upf-toastr--container');
+      if (container?.children.length === 0) {
+        container.remove();
+      }
+    }
+  }
+
+  private _onToastClose(evt: Event) {
```
Miexil, line 107: `:void`

Miexil (changes requested):

addon/services/toast.ts
```diff
@@ -0,0 +1,129 @@
...
+    messageContainer.textContent = message;
+
+    let toastChildren: Element[] = [titleContainer, messageContainer];
+
+    if (opts.closeButton) {
+      const closeButton: Element = document.createElement('button');
+
+      closeButton.classList.add('toast-close-button');
+      closeButton.innerHTML = '<i class="fas fa-times text-size-5"></i>';
+      closeButton.addEventListener('click', this._onToastClose.bind(this), { once: true });
```
phndiaye, line 65: That `once` parameter is going to change my life 😅 

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#parameters / https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#browser_compatibility

Miexil: Nice ! 👍

addon/services/toast.ts
```diff
@@ -0,0 +1,129 @@
...
+
+export default class Toast extends Service {
+  private _toasts: Element[] = [];
+
+  success: ToastFunction = delegate(this, ToastType.SUCCESS);
+  error: ToastFunction = delegate(this, ToastType.ERROR);
+  warning: ToastFunction = delegate(this, ToastType.WARNING);
+  info: ToastFunction = delegate(this, ToastType.INFO);
+
+  buildToast(type: ToastType, message: string, title: string, opts: ToastOptions): void {
```
Miexil, line 44: One last thing (apart from tests ! 😂 )
This method is a bit too long and can be split a bit more.

For instance, you create titleContainer & messageContainer, then apply the same logic underneath both of them. Why not have a createElementWithClassAndText method ? (rename it ofc ^^)

The closeButton logic in the if() can also be moved to a separate method.

addon/services/toast.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    toastChildren.forEach((x) => toast.append(x));
+
+    if (container.querySelector('.toast')) {
+      container.insertBefore(toast, container.querySelector('.toast'));
+    } else {
+      container.append(toast);
+    }
+
+    later(
```
Miexil, line 82: Can you explain why this is run in a timeout/later ?
Also, you might want to put that 500 timeout into a constant to make it clearer :)

addon/services/toast.ts
JulienVannier66: Wow!

tests/dummy/app/styles/app.less
phndiaye: all as he said :p

JulienVannier66 (approved):
LGTM

addon/services/toast.ts
```diff
@@ -0,0 +1,185 @@
...
+      if (isEmpty(message) && isEmpty(title)) return;
+
+      this._buildToast(type, message, title, { ...DEFAULT_OPTIONS, ...opts });
+    };
+  }
+
+  private _buildToast(type: ToastType, message: string, title: string, opts: ToastOptions): void {
+    const container: Element = this._buildContainer();
+
+    const toast: Element = document.createElement('div');
```
Miexil, line 93: use `_buildToastPart` ?

addon/services/toast.ts
```diff
@@ -0,0 +1,185 @@
...
+    toast.classList.add('toast', 'toast--hidden', `toast-${type}`);
+
+    const titleContainer: Element = this._buildToastPart('div', ['toast-title'], title);
+    const messageContainer: Element = this._buildToastPart('div', ['toast-message'], message);
+
+    let toastChildren: Element[] = [titleContainer, messageContainer];
+
+    if (opts.closeButton) {
```
Miexil, line 101: I feel this `_buildToast` method is still too long.
this `closeButton` logic can be moved to a dedicated method

addon/services/toast.ts
```diff
@@ -0,0 +1,185 @@
...
+    toast.classList.remove('toast--hidden');
+    toast.classList.add('toast--visible');
+
+    this._toasts.push(toast);
+
+    if (opts.timeout === 'none') {
+      return;
+    }
+
+    later(this, () => this._destroyToast(toast), opts.timeout || DEFAULT_TOAST_TIMEOUT);
```
Miexil, line 139: Move this logic into a separate method ?
`handleToastDisplay`, `displayToast` ? (idk 🤷 )

addon/services/toast.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  private _buildContainer(): Element {
+    let container: Element | null = document.querySelector('body #toast-container.upf-toastr--container');
+
+    if (!container) {
+      container = document.createElement('div');
+      container.id = 'toast-container';
+      container.classList.add('upf-toastr--container');
+
```
Miexil, line 172: random line break 🙂

addon/services/toast.ts
phndiaye: @Miexil can't split much i think ahah

Miexil (approved):
LGTM, well done 👍

PR #131, opened 2021-11-16
Add inline args to disable floating menu and translation

JulienVannier66:
### What does this PR do?

Add inline arg to disable floating menu and translation

Related to: https://github.com/upfluence/backlog/issues/1107

Screenshot from 2021-11-16 16-38-57
Peek 2021-11-16 16-38
### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/infinite-select.hbs
+++ b/addon/components/o-s-s/infinite-select.hbs
@@ -33,12 +33,12 @@
 
           <div class="margin-top-xx-sm text-color-default upf-align--center">
             {{#if this._searchKeyword}}
-              <p>No matching results</p>
+              <p>{{t "oss-components.infinite-select.no-match.title"}}</p>
               <p class="text-color-default-light">
-                Try adjusting your search to find what you’re looking for.
+                {{t "oss-components.infinite-select.no-match.description"}}
               </p>
             {{else}}
-              Nothing to see here.
+              {{t "oss-components.infinite-select.empty"}}
             {{/if}}
           </div>
         </div>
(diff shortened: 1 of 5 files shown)


addon/components/o-s-s/infinite-select.hbs
```diff
@@ -1,4 +1,4 @@
-<div class="upf-infinite-select" {{did-insert this.onRender}} ...attributes>
+<div class="fx-col upf-infinite-select {{if (not this.inline) 'upf-infinite-select--absolute'}}" {{did-insert this.onRender}} ...attributes>
```
phndiaye, line 1: isn't it better to do `if inline -> upf-infinite-select--inline` to:
1) avoid the negations
2) don't break the existing because right now, if we bump oss independently, it will break existing :( 
3) keep the current version as default

tests/integration/components/o-s-s/infinite-select-test.js
```diff
@@ -200,5 +200,39 @@ module('Integration | Component | o-s-s/infinite-select', function (hooks) {
...
+        this.items = FAKE_DATA;
+        this.onSelect = () => {};
+
+        await render(
+          hbs`<OSS::InfiniteSelect @items={{this.items}} @inline={{false}} @searchEnabled={{false}} @onSelect={{this.onSelect}}/>`
+        );
+
+        assert.dom('.upf-infinite-select.upf-infinite-select--absolute').exists({ count: 1 });
+      });
+      test('it should render with truly arg', async function (assert) {
```
Miexil, line 225: Truthy *

Miexil (approved):

phndiaye (approved):

PR #132, opened 2021-11-17
Add OSS::Link component and add args to OSS::Button & OSS::InputContainer

phndiaye:
### What does this PR do?

- Adds a new component: OSS::Link for displaying transparent "buttons" that trigger actions.
- Adds `theme` argument to the OSS::Button component to control its display on dark backgrounds.
- Adds `disabled`. argument to OSS::InputContainer to disable the inner input easily without having having to pass an `input` named block.

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/link.hbs
+++ b/addon/components/o-s-s/link.hbs
@@ -1,0 +1,9 @@
+<a onclick="event.preventDefault()" href="" class="upf-link" ...attributes>
+  {{#if @icon}}
+    <i class={{@icon}}></i>
+  {{/if}}
+
+  {{#if @label}}
+    <span class={{if @icon "margin-left-xxx-sm"}}>{{@label}}</span>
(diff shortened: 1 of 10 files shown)


Miexil (approved):

addon/components/o-s-s/link.hbs
```diff
@@ -0,0 +1,9 @@
+<a onclick="event.preventDefault()" href="" class="upf-link" ...attributes>
```
JulienVannier66, line 1: What is the best option:
- use native js like in the code
- use `{{on 'click' this.preventDefault}}` and declare the action in ts file

I don't think that put functional logic in `hbs` is a good idea for the readability

JulienVannier66 (approved):

PR #134, opened 2021-11-22
Fix tests on OSS:PowerSelect

JulienVannier66:
### What does this PR do?

Fix flaky tests of power-select

Related to: https://github.com/upfluence/backlog/issues/1107

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/power-select.hbs
+++ b/addon/components/o-s-s/power-select.hbs
@@ -2,6 +2,7 @@
      {{did-insert (fn this.ensureBlockPresence (has-block 'selected-item') (has-block 'option-item'))}}>
   <div class="upf-power-select__array-container margin-bottom-xxx-sm">
     <div class="array-input-container {{if @selectedItems 'fx-row fx-xalign-center' 'fx-col fx-malign-center'}} {{if this.hasScrollbar 'has-overflow'}}"
+         data-control-name="power-select_array-input-container"
          {{did-insert this.observeResize}} {{will-destroy this.unobserveResize}}>
       {{#each @selectedItems as |selectedItem|}}
         <div class="input-array-tag fx-row fx-xalign-center">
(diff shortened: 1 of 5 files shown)


phndiaye (approved):

tests/integration/components/o-s-s/power-select-test.ts
```diff
@@ -242,9 +243,14 @@ module('Integration | Component | o-s-s/power-select', function (hooks) {
         </div>
       `);
 
-      await scrollTo('.upf-infinite-select__items-container', 0, 100);
+      await scrollTo(
+        '.upf-infinite-select__items-container',
+        0,
+        document.querySelector('.upf-infinite-select__items-container')!.scrollHeight -
+          document.querySelector('.upf-infinite-select__items-container')!.clientHeight
+      );
```
Miexil, line 251: FYI, inputing a high value (such as 1500) works as well.

tests/integration/components/o-s-s/power-select-test.ts
```diff
@@ -242,9 +243,14 @@ module('Integration | Component | o-s-s/power-select', function (hooks) {
...
 
-      assert.ok(this.onBottomReached.calledOnce);
```
Miexil, line 247: `calledOnce` should've worked, if it didn't it means that it is called mutliple times, which is weird no ? 🤔

JulienVannier66: The error maybe come from Infinite-select?

addon/components/o-s-s/power-select.hbs
```diff
@@ -2,6 +2,7 @@
      {{did-insert (fn this.ensureBlockPresence (has-block 'selected-item') (has-block 'option-item'))}}>
   <div class="upf-power-select__array-container margin-bottom-xxx-sm">
     <div class="array-input-container {{if @selectedItems 'fx-row fx-xalign-center' 'fx-col fx-malign-center'}} {{if this.hasScrollbar 'has-overflow'}}"
+         data-control-name="power-select_array-input-container"
```
phndiaye, line 5: not sure it's the role of the power-select itself to set this as it depends on the context it's used in (it will be passed in the top `...attributes` by the parent application.

Miexil (approved):

phndiaye (approved):

PR #141, opened 2021-12-08
use font awesome pro's fonts if present in the parent's deps

phndiaye:
### What does this PR do?

Use font awesome pro when present.

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,6 +27,7 @@ module.exports = {
 
   included: function (app) {
     this._super.included.apply(this, app);
+    this.hasFontAwesomePro = Object.keys(app.dependencies()).includes('@fortawesome/fontawesome-pro');
 
     this._registerLessDependencies(app);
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
 
   included: function (app) {
     this._super.included.apply(this, app);
+    this.hasFontAwesomePro = Object.keys(app.dependencies()).includes('@fortawesome/fontawesome-pro');
```
Miexil, line 30: FortAwesome is not a typo right ? 🙂

phndiaye: nope ! never understood 😅

Miexil (approved):

JulienVannier66 (approved):

PR #142, opened 2021-12-10
Add isTesting for modals

JulienVannier66:
### What does this PR do?

Add `isTesting` for OSS::Modal to fix tests
Add font awesome pro

Related to: nothing

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/modal.js
+++ b/addon/components/o-s-s/modal.js
@@ -1,6 +1,7 @@
 import Component from '@glimmer/component';
 import { action } from '@ember/object';
 import jQuery from 'jquery';
+import { isTesting } from '@embroider/macros';
 
 const DEFAULT_OPTIONS = {
   centered: false,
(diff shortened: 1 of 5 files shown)


addon/components/o-s-s/modal.js
```diff
@@ -18,6 +19,10 @@ export default class OssModalComponent extends Component {
     return { ...DEFAULT_OPTIONS, ...(this.args.options || {}) };
   }
 
+  get container() {
+    return isTesting() ? null : this.options.container;
```
Miexil, line 23: I think we tried but I can't remember the result. What happens when running tests if we set the 'ember-testing-container' instead of null here ?

JulienVannier66: It works, but if I set a random value, it works too. It's weird...

phndiaye: the backdrop was still in the body, also

phndiaye (approved):

Miexil (approved):

PR #146, opened 2021-12-20
Remove all linter issues

Miexil:
### What does this PR do?

Fixes all linter issues that were raised.
Adds a Makefile with a set of commands (same as on publishr-admin) and some scripts.

The goal after this is to enable the Linting & Tests steps into the CI.

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled


tests/dummy/app/templates/application.hbs
```diff
@@ -43,16 +43,28 @@
...
+                       @ongoingMessage="ongoing"
+                       @successMessage="success"
+                       @failureMessage="failure'"
+                       class="margin-right-xx-sm">
+      Destructive button
+    </DestructiveButton>
+    <LoadingButton @slowAction="loadingAction" class="upf-btn upf-btn--primary"
+                   @initiallyDisabled={{false}}>
+      Loading button
+    </LoadingButton>
```
JulienVannier66, line 68: Do you keep that and these functions associated?

Miexil: Yeah, I think that's fine ; we've been leaving component examples together with @phndiaye.
It's the test app so nothing problematic IMHO.

JulienVannier66 (approved):

phndiaye (approved):

tests/dummy/app/templates/application.hbs
phndiaye: yep, it's only ran locally so doesn't do any harm

phndiaye (approved):

JulienVannier66 (approved):

PR #151, opened 2022-01-10
Add new OSS::PhoneNumberInput component

Miexil:
### What does this PR do?

Adds a new PhoneNumberInput component.

Related to: [#1269](https://github.com/upfluence/backlog/issues/1269)

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled


app/styles/phone-number-input.less
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  .phone-number-input {
+    .country-selector {
+      border-right: 1px solid @upf-gray;
+      height: 100%;
+      display: flex;
+      align-items: center;
+      padding-right: 12px;
+      margin-right: 12px;
```
JulienVannier66, line 26:
```suggestion
      padding-right: @spacing-xx-sm;
      margin-right: @spacing-xx-sm;
```

addon/components/o-s-s/phone-number-input.ts
```diff
@@ -0,0 +1,81 @@
+import { action } from '@ember/object';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { countries, CountryData } from '@upfluence/oss-components/utils/country-codes';
+
+interface OSSPhoneNumberInputArgs {
+  prefix: string;
+  number: string;
+  onChange: (prefix: string, number: string) => void;
```
JulienVannier66, line 9:
```suggestion
  onChange(prefix: string, number: string): void;
```
We don't use this syntax now?

app/styles/phone-number-input.less
Miexil: That's not the same width sadly, I tried using our variables but couldn't get something that matched exactly.

addon/components/o-s-s/phone-number-input.ts
Miexil: We sure do !

app/styles/phone-number-input.less
JulienVannier66: Ok ok no problem, it's good for me so

JulienVannier66 (approved):

addon/components/o-s-s/phone-number-input.hbs
```diff
@@ -0,0 +1,32 @@
+<div class="phone-number-container fx-1" ...attributes>
+  <div class="phone-number-input upf-input fx-row fx-1 fx-xalign-center">
+    <div class="country-selector fx-row" role="button" {{on "click" this.toggleCountrySelector}}>
+      <div class="fflag fflag-{{this.selectedCountry.id}} ff-sm ff-wave ff-rounded"></div>
+      {{#if this.countrySelectorShown}}
+        <i class="fal fa-chevron-up margin-left-xxx-sm"></i>
```
phndiaye, line 6: so we start using the `fal` ?

Miexil: In this case it is just a thinner chevron ; is that an issue ? 🤔

phndiaye (approved):

PR #172, opened 2022-05-27
Add stop-propagation helper

phndiaye:
### What does this PR do?

Because it's something we have to do a lot, this helper will allow us to invoke `event.stopPropagation` directly from the templates.

Whilst this can be useless if we have an action defined in the component JS where we can handle the event, it can be useful when mutating props with `mut` or calling a parent-provided action:
-  `{{on "click" (fn (mut this.myProp) false)}}` ->  `{{on "click" (stop-propagation (fn (mut this.myProp) false))}}`
-  `{{on "click" @parentAction}}` -> `{{on "click" (stop-propagation @parentAction)}}`

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/tests/integration/helpers/stop-propagation-test.ts
+++ b/tests/integration/helpers/stop-propagation-test.ts
@@ -1,0 +1,31 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { click, render } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+
+module('Integration | Helper | stop-propagation', function (hooks) {
+  setupRenderingTest(hooks);
+
+  hooks.beforeEach(function () {
+    this.nativeStopPropagation = Event.prototype.stopPropagation;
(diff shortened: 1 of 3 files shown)


Sypheos (approved):

tests/integration/helpers/stop-propagation-test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  hooks.beforeEach(function () {
+    this.nativeStopPropagation = Event.prototype.stopPropagation;
+  });
+
+  hooks.afterEach(function () {
+    Event.prototype.stopPropagation = this.nativeStopPropagation;
+  });
+
+  // Replace this with your real tests.
```
Miexil, line 17: Remove comment 😇

Miexil (approved):
LGTM except for the comment line.

JulienVannier66 (approved):

PR #173, opened 2022-05-31
New OSS::Alert component

JulienVannier66:
### What does this PR do?

- Add OSS::Alert component & tests & doc
- Fix toast text size
- Clean up &nbsp;

Related to: https://github.com/upfluence/backlog/issues/1667

Peek 2022-05-31 15-19

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled


addon/components/o-s-s/alert.ts
```diff
@@ -0,0 +1,38 @@
...
+  subtitle?: string;
+}
+
+export default class OSSAlert extends Component<OSSAlertArgs> {
+  constructor(owner: unknown, args: OSSAlertArgs) {
+    super(owner, args);
+
+    if (!args.skin) {
+      throw new Error('[component][OSS::Alert] The @skin argument is mandatory');
+    }
```
Miexil, line 17: how about setting 'info' as default ? 🤷

Miexil: You could set the default state of your switch below to 'fa-info-circle' & remove the `case 'info'` line ?

app/styles/alert.less
```diff
@@ -0,0 +1,81 @@
...
+    padding: @spacing-xx-sm;
+    gap: @spacing-xx-sm;
+    background-color: var(--color-gray-50);
+  }
+
+  .text-container {
+    display: flex;
+    flex-direction: column;
+    flex: 1;
+    gap: 3px;
```
Miexil, line 22: This is to "skip" the position of the "indicator" element right ?
i suggest you define a local variable for this ; it would make more sense while reading the code.

tests/integration/components/o-s-s/alert-test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  test('it throws an error if no skin argument is passed', async function (assert) {
+    setupOnerror((err: any) => {
+      assert.equal(err.message, '[component][OSS::Alert] The @skin argument is mandatory');
+    });
+
+    await render(hbs`<OSS::Alert />`);
+  });
+
+  ALERT_SKINS.forEach((skin) => {
```
Miexil, line 31: 👍

addon/components/o-s-s/alert.hbs
```diff
@@ -0,0 +1,16 @@
+<div class={{concat "upf-alert-v2 " this.skinClass}} ...attributes>
+  <div class="indicator"></div>
+  <div class="main-container">
+    <span class="icon">
```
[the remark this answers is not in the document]
JulienVannier66: Yes, if I remove it, the icon is not align with the title

addon/components/o-s-s/alert.ts
JulienVannier66: Ok ok I make 'info' as default state

app/styles/alert.less
JulienVannier66: No, this is just the gap between title and subtile. I can set it in variable, no problem with that

PR #180, opened 2022-06-16
OSS::CountrySelector initial value triggers onChange + tests

Miexil:
### What does this PR do?

OSS::CountrySelector initial value triggers onChange + tests
Related to: #[1338](https://github.com/upfluence/backlog/issues/1338)

@initialValue triggers the onChange method to notify there is a match to the parent.
Tests updated & tests added for initialValue

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/tests/integration/components/o-s-s/country-selector-test.ts
+++ b/tests/integration/components/o-s-s/country-selector-test.ts
@@ -62,29 +62,30 @@ module('Integration | Component | o-s-s/country-selector', function (hooks) {
     assert.dom('.upf-infinite-select').exists();
   });
 
-  module('If @initialValue is passed', () => {
+  module('If @initialValue is passed', function (hooks) {
+    hooks.beforeEach(function () {
+      this.initValue = {
+        id: 'FR',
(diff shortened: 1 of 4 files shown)


tests/integration/components/o-s-s/country-selector-test.ts
```diff
@@ -62,6 +62,32 @@ module('Integration | Component | o-s-s/country-selector', function (hooks) {
     assert.dom('.upf-infinite-select').exists();
   });
 
+  module('If @initialValue is passed', () => {
+    test('If the value matches an entry from the sourceList, then the input is set to the value', async function (assert) {
+      await render(
+        hbs`<OSS::CountrySelector @onChange={{this.onchange}} @sourceList={{this.countries}} @initialValue="France" />`
```
phndiaye, line 68: what about using the iso-codes for initial value ?

JulienVannier66 (approved):

PR #189, opened 2022-06-30
Use assert to check for mandatory args

phndiaye:
### What does this PR do?

Use assert to check for mandatory args

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/modal.stories.js
+++ b/addon/components/o-s-s/modal.stories.js
@@ -5,7 +5,7 @@ export default {
   title: 'Components/OSS::Modal',
   component: 'modal',
   args: {
-    title: 'Moodal Title'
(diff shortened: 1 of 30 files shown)


addon/components/o-s-s/modal.stories.js
```diff
@@ -5,7 +5,7 @@ export default {
   title: 'Components/OSS::Modal',
   component: 'modal',
   args: {
-    title: 'Moodal Title'
```
JulienVannier66, line 8: Moodal :laughing:

JulienVannier66 (approved):
LGTM

PR #191, opened 2022-07-07
Add param on OSS::CurrencyInput for show only the currency dd

sBenjamin270495:
### What does this PR do?

Add param on OSS::CurrencyInput for show only the currency dropdown without number input

Related to: none, advancement for replace in identity web

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/currency-input.hbs
+++ b/addon/components/o-s-s/currency-input.hbs
@@ -25,7 +25,7 @@
   {{#if this.currencySelectorShown}}
     <OSS::InfiniteSelect @items={{this.filteredCurrencies}} @onSearch={{this.onSearch}} 
                          @onSelect={{this.onSelect}}
-                         @searchPlaceholder={{t "oss-components.currency-input.search" }} 
+                         @searchPlaceholder={{t "oss-components.currency-input.search"}} 
                          {{on-click-outside this.hideCurrencySelector}}>
       <:option as |currency|>
         <div class="fx-row fx-xalign-center {{if (eq this.selectedCurrency currency) 'row-selected'}}">
--- a/addon/components/o-s-s/currency-input.hbs
+++ b/addon/components/o-s-s/currency-input.hbs
@@ -2,40 +2,40 @@
   <div class="currency-input {{if @onlyCurrency " onlycurrency"}} upf-input fx-row fx-1 fx-xalign-center">
     <div class="currency-selector fx-row fx-gap-px-12 fx-malign-space-between" role="button" 
          {{on "click" this.toggleCurrencySelector}}>
+    <div class="currency-selector fx-row fx-gap-px-12 fx-malign-space-between" role="button" {{on "click"
+      this.toggleCurrencySelector}}>
       <div class="fx-col">
         <div class="fx-row fx-gap-px-9">
           <span>{{this.selectedCurrencySymbol}}</span>
           {{#if @onlyCurrency}}
-            <span class="margin-right-px-12">{{this.selectedCurrencyCode}}</span>
+          <span class="margin-right-px-12">{{this.selectedCurrencyCode}}</span>
           {{/if}}
         </div>
       </div>
       {{#if this.currencySelectorShown}}
-        <i class="far fa-chevron-up margin-left-xxx-sm"></i>
+      <i class="far fa-chevron-up margin-left-xxx-sm"></i>
       {{else}}
-        <i class="far fa-chevron-down margin-left-xxx-sm"></i>
+      <i class="far fa-chevron-down margin-left-xxx-sm"></i>
       {{/if}}
     </div>
     {{#unless @onlyCurrency}}
-      <Input class="fx-1" type="number" @value={{this.localValue}} {{on "keydown" this.onlyNumeric}}
-             {{on "keyup" this.notifyChanges}}
-             autocomplete="off" />
+    <Input class="fx-1" type="number" @value={{this.localValue}} {{on "keydown" this.onlyNumeric}} {{on "keyup"
+      this.notifyChanges}} autocomplete="off" />
     {{/unless}}
   </div>
   {{#if this.currencySelectorShown}}
-    <OSS::InfiniteSelect @items={{this.filteredCurrencies}} @onSearch={{this.onSearch}} 
-                         @onSelect={{this.onSelect}}
-                         @searchPlaceholder={{t "oss-components.currency-input.search"}} 
-                         {{on-click-outside this.hideCurrencySelector}}>
-      <:option as |currency|>
-        <div class="fx-row fx-xalign-center {{if (eq this.selectedCurrency currency) 'row-selected'}}">
-          <span class="symbol text-color-default-light margin-left-xx-sm">{{currency.symbol}}</span>
-          <span class="text-color-default-light margin-left-xx-sm fx-1">{{currency.code}}</span>
-          {{#if (eq this.selectedCurrency currency)}}
-            <i class="far fa-check text-color-bright-purple padding-right-px-6"></i>
-          {{/if}}
-        </div>
-      </:option>
-    </OSS::InfiniteSelect>
+  <OSS::InfiniteSelect @items={{this.filteredCurrencies}} 
+                       @onSearch={{this.onSearch}} @onSelect={{this.onSelect}}
+    @searchPlaceholder={{t "oss-components.currency-input.search" }} {{on-click-outside this.hideCurrencySelector}}>
+    <:option as |currency|>
+      <div class="fx-row fx-xalign-center {{if (eq this.selectedCurrency currency) 'row-selected'}}">
+        <span class="symbol text-color-default-light margin-left-xx-sm">{{currency.symbol}}</span>
+        <span class="text-color-default-light margin-left-xx-sm fx-1">{{currency.code}}</span>
+        {{#if (eq this.selectedCurrency currency)}}
+        <i class="far fa-check text-color-bright-purple padding-right-px-6"></i>
+        {{/if}}
+      </div>
+    </:option>
+  </OSS::InfiniteSelect>
   {{/if}}
 </div>
--- a/addon/components/o-s-s/currency-input.hbs
+++ b/addon/components/o-s-s/currency-input.hbs
@@ -1,34 +1,31 @@
 <div class="currency-input-container fx-1" ...attributes>
-  <div class="{{if this.onlyCurrencyInput "currency-input-onlycurrency" "currency-input"}} upf-input fx-row fx-1 fx-xalign-center ">
-    <div class="currency-selector fx-row fx-malign-space-between" role="button" {{on "click" this.toggleCurrencySelector}}>
+  <div class="currency-input {{if @onlyCurrency " onlycurrency"}} upf-input fx-row fx-1 fx-xalign-center">
+    <div class="currency-selector fx-row fx-gap-px-12 fx-malign-space-between" role="button" {{on "click"
+      this.toggleCurrencySelector}}>
       <div class="fx-col">
-        <div class="fx-row fx-gap-px-24">
-          <span class="text-center symbol">{{this.selectedCurrencySymbol}}</span>
-          {{#if this.onlyCurrencyInput}}
-            <span class=" margin-right-px-24 text-center symbol">{{this.selectedCurrencyCode}}</span>
+        <div class="fx-row fx-gap-px-9">
+          {{#if (not-eq this.selectedCurrencyCode "Select")}}
+            <span>{{this.selectedCurrencySymbol}}</span>
+          {{/if}}
+          {{#if @onlyCurrency}}
+            <span class=" margin-right-px-12 {{if (eq this.selectedCurrencyCode "Select") "font-color-gray-500" }}">{{this.selectedCurrencyCode}}</span>
           {{/if}}
         </div>
       </div>
-      <div class="fx-col">
       {{#if this.currencySelectorShown}}
         <i class="far fa-chevron-up margin-left-xxx-sm"></i>
       {{else}}
         <i class="far fa-chevron-down margin-left-xxx-sm"></i>
       {{/if}}
-
-      </div>
     </div>
-    {{#unless this.onlyCurrencyInput}}
-      <Input class="fx-1" type="number" @value={{this.localValue}} {{on "keydown" this.onlyNumeric}}
-             {{on "keyup" this.notifyChanges}}
-             autocomplete="off" />
+    {{#unless @onlyCurrency}}
+      <Input class="fx-1" type="number" @value={{this.localValue}} {{on "keydown" this.onlyNumeric}} {{on "keyup"
+        this.notifyChanges}} autocomplete="off" />
     {{/unless}}
   </div>
   {{#if this.currencySelectorShown}}
-    <OSS::InfiniteSelect @items={{this.filteredCurrencies}} @onSearch={{this.onSearch}}
-                         @onSelect={{this.onSelect}}
-                         @searchPlaceholder={{t "oss-components.currency-input.search"}}
-                         {{on-click-outside this.hideCurrencySelector}}>
+    <OSS::InfiniteSelect @items={{this.filteredCurrencies}} @onSearch={{this.onSearch}} @onSelect={{this.onSelect}}
+      @searchPlaceholder={{t "oss-components.currency-input.search" }} {{on-click-outside this.hideCurrencySelector}}>
       <:option as |currency|>
         <div class="fx-row fx-xalign-center {{if (eq this.selectedCurrency currency) 'row-selected'}}">
           <span class="symbol text-color-default-light margin-left-xx-sm">{{currency.symbol}}</span>
--- a/addon/components/o-s-s/currency-input.hbs
+++ b/addon/components/o-s-s/currency-input.hbs
@@ -2,8 +2,12 @@
   <div class="{{if this.onlyCurrencyInput "currency-input-onlycurrency" "currency-input"}} upf-input fx-row fx-1 fx-xalign-center ">
     <div class="currency-selector fx-row fx-malign-space-between" role="button" {{on "click" this.toggleCurrencySelector}}>
       <div class="fx-col">
-        <span class="margin-right-xxx-sm text-center symbol">{{this.selectedCurrencySymbol}}</span>
-
+        <div class="fx-row fx-gap-px-24">
+          <span class="text-center symbol">{{this.selectedCurrencySymbol}}</span>
+          {{#if this.onlyCurrencyInput}}
+            <span class=" margin-right-px-24 text-center symbol">{{this.selectedCurrencyCode}}</span>
+          {{/if}}
+        </div>
       </div>
       <div class="fx-col">
       {{#if this.currencySelectorShown}}
--- a/addon/components/o-s-s/currency-input.hbs
+++ b/addon/components/o-s-s/currency-input.hbs
@@ -1,14 +1,12 @@
 <div class="currency-input-container fx-1" ...attributes>
   <div class="currency-input {{if @onlyCurrency " onlycurrency"}} upf-input fx-row fx-1 fx-xalign-center">
-    <div class="currency-selector fx-row fx-gap-px-12 fx-malign-space-between" role="button" {{on "click"
-      this.toggleCurrencySelector}}>
+    <div class="currency-selector fx-row fx-gap-px-12 fx-malign-space-between" role="button" 
+         {{on "click" this.toggleCurrencySelector}}>
       <div class="fx-col">
         <div class="fx-row fx-gap-px-9">
-          {{#if (not-eq this.selectedCurrencyCode "Select")}}
-            <span>{{this.selectedCurrencySymbol}}</span>
-          {{/if}}
+          <span>{{this.selectedCurrencySymbol}}</span>
           {{#if @onlyCurrency}}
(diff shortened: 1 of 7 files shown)


addon/components/o-s-s/currency-input.hbs
```diff
@@ -1,16 +1,28 @@
 <div class="currency-input-container fx-1" ...attributes>
-  <div class="currency-input upf-input fx-row fx-1 fx-xalign-center">
-    <div class="currency-selector fx-row" role="button" {{on "click" this.toggleCurrencySelector}}>
-      <span class="margin-right-xxx-sm text-center symbol">{{this.selectedCurrencySymbol}}</span>
+  <div class="{{if this.onlyCurrencyInput "currency-input-onlycurrency" "currency-input"}} upf-input fx-row fx-1 fx-xalign-center ">
```
JulienVannier66, line 2:
```suggestion
  <div class="currency-input {{if this.onlyCurrencyInput "currency-input--currency-only"}} upf-input fx-row fx-1 fx-xalign-center ">
```
Keep `currency-input` class to avoid code duplication and update `currency-input--currency-only` to change only necessary css properties

app/styles/currency-input.less
```diff
@@ -47,6 +47,37 @@
...
+      padding-right: var(--spacing-px-12);
+      margin-right: 0px;
+      width: 100%;
+    }
+
+    .phone-prefix {
+      white-space: nowrap;
+      .text-color-default-light;
+      .margin-right-xxx-sm;
+    }
```
Miexil, line 69: `.phone-prefix` is not used in currency-input.
Be careful when copy/pasting code :)

app/styles/currency-input.less
```diff
@@ -47,6 +47,37 @@
...
+  .currency-input-onlycurrency {
+    padding-right: 0px;
+    .currency-selector {
+      height: 100%;
+      display: flex;
+      align-items: center;
+      padding-right: var(--spacing-px-12);
+      margin-right: 0px;
+      width: 100%;
+    }
```
Miexil, line 63: There are only a few css attributes that are different from the base `.currency-input` : width, margin, border...
Instead of duplicating the whole section, define a class that only overrides the attributes you need.

In your template you should do something like :
```html
 <div class="currency-input {{if this.onlyCurrencyInput "only-selector"}} ..." ...>
```
And then in the css, override the currency-input attributes that are different in `.only-selector`

addon/components/o-s-s/currency-input.hbs
Miexil, line 2: extra space at the end of class definitions.

addon/components/o-s-s/currency-input.hbs
```diff
@@ -1,16 +1,28 @@
...
+    <div class="currency-selector fx-row fx-malign-space-between" role="button" {{on "click" this.toggleCurrencySelector}}>
+      <div class="fx-col">
+        <div class="fx-row fx-gap-px-24">
+          <span class="text-center symbol">{{this.selectedCurrencySymbol}}</span>
+          {{#if this.onlyCurrencyInput}}
+            <span class=" margin-right-px-24 text-center symbol">{{this.selectedCurrencyCode}}</span>
+          {{/if}}
```
JulienVannier66, line 9: You see that with Victor, right?

addon/components/o-s-s/currency-input.hbs
```diff
@@ -1,16 +1,28 @@
...
+        </div>
+      </div>
+      <div class="fx-col">
```
Miexil, line 12: This feels a bit off. Why are you adding these divs ?

Why are you using `text-center` & `symbol` and what's `symbol` ?

addon/components/o-s-s/currency-input.hbs
```diff
@@ -1,16 +1,28 @@
...
+      </div>
     </div>
-    <Input class="fx-1" type="number" @value={{this.localValue}} {{on "keydown" this.onlyNumeric}}
-           {{on "keyup" this.notifyChanges}}
-           autocomplete="off" />
+    {{#unless this.onlyCurrencyInput}}
+      <Input class="fx-1" type="number" @value={{this.localValue}} {{on "keydown" this.onlyNumeric}}
+             {{on "keyup" this.notifyChanges}}
+             autocomplete="off" />
+    {{/unless}}
```
Miexil, line 25: indentation of the whole file is wrong.

addon/components/o-s-s/currency-input.ts
```diff
@@ -31,6 +32,14 @@ export default class OSSCurrencyInput extends Component<OSSCurrencyInputArgs> {
     return this.selectedCurrency.symbol || '—';
   }
 
+  get selectedCurrencyCode(): string {
+    return this.selectedCurrency.code || '...';
```
Miexil, line 36: I doubt that we put the '...' when no value is selected ; please check with Victor.

app/styles/currency-input.less
```diff
@@ -47,6 +47,37 @@
...
+
+    .phone-prefix {
+      white-space: nowrap;
+      .text-color-default-light;
+      .margin-right-xxx-sm;
+    }
+
+    input {
```
Miexil, line 71: pretty sure you don't need to override inputs

tests/integration/components/o-s-s/currency-input-test.ts
```diff
@@ -16,6 +16,23 @@ module('Integration | Component | o-s-s/currency-input', function (hooks) {
...
+    assert.dom('.currency-input-container').exists();
+    assert.dom('.currency-selector ').exists();
+
+    const currencyLabel = await findAll('.currency-selector span');
+    assert.dom(currencyLabel[1]).exists();
+    assert.dom(currencyLabel[1]).hasText('USD');
+
+    assert.dom('.currency-input input ').doesNotExist();
+  });
+
```
Miexil, line 35: You should also write a test to check that onChange is called with the proper values when currencyOnly is true

Miexil (changes requested):

JulienVannier66:
Do you forget the `min-width: 150px` for `.currency-input-container .upf-infinite-select` class? The renderer is weird

addon/components/o-s-s/currency-input.ts
```diff
@@ -8,6 +8,7 @@ interface OSSCurrencyInputArgs {
   currency: string;
   value: number;
   onChange(currency: string, value: number): void;
+  onlyCurrency: boolean;
```
phndiaye, line 11: `currencyOnly?: boolean`.

In Typescript, the `?` at the end means that the parameter can be undefined and is a shortcut for:
- `onlyCurrency: boolean | undefined`

addon/components/o-s-s/currency-input.hbs
```diff
@@ -1,16 +1,28 @@
 <div class="currency-input-container fx-1" ...attributes>
-  <div class="currency-input upf-input fx-row fx-1 fx-xalign-center">
-    <div class="currency-selector fx-row" role="button" {{on "click" this.toggleCurrencySelector}}>
-      <span class="margin-right-xxx-sm text-center symbol">{{this.selectedCurrencySymbol}}</span>
+  <div class="{{if this.onlyCurrencyInput "currency-input-onlycurrency" "currency-input"}} upf-input fx-row fx-1 fx-xalign-center ">
+    <div class="currency-selector fx-row fx-malign-space-between" role="button" {{on "click" this.toggleCurrencySelector}}>
+      <div class="fx-col">
+        <div class="fx-row fx-gap-px-24">
+          <span class="text-center symbol">{{this.selectedCurrencySymbol}}</span>
+          {{#if this.onlyCurrencyInput}}
```
phndiaye, line 7: just use the `@currencyOnly` named arg and remove this getter (cf. https://github.com/upfluence/oss-components/pull/191/files#r915987263)

addon/components/o-s-s/currency-input.hbs
```diff
@@ -1,16 +1,28 @@
...
+            <span class=" margin-right-px-24 text-center symbol">{{this.selectedCurrencyCode}}</span>
```
phndiaye, line 8: you really want to apply the `symbol` class on the currency code too ??

app/styles/currency-input.less
sBenjamin270495: the input class is needed for border : none  on the input , doesn't have same design without this class

Miexil (changes requested):

addon/components/o-s-s/currency-input.hbs
```diff
@@ -1,22 +1,31 @@
 <div class="currency-input-container fx-1" ...attributes>
-  <div class="currency-input upf-input fx-row fx-1 fx-xalign-center">
-    <div class="currency-selector fx-row" role="button" {{on "click" this.toggleCurrencySelector}}>
-      <span class="margin-right-xxx-sm text-center symbol">{{this.selectedCurrencySymbol}}</span>
+  <div class="currency-input {{if @onlyCurrency " onlycurrency"}} upf-input fx-row fx-1 fx-xalign-center">
+    <div class="currency-selector fx-row fx-gap-px-12 fx-malign-space-between" role="button" {{on "click"
+      this.toggleCurrencySelector}}>
+      <div class="fx-col">
+        <div class="fx-row fx-gap-px-9">
+          {{#if (not-eq this.selectedCurrencyCode "Select")}}
```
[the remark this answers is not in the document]
phndiaye: Also, "Select" IS NOT a currency code. let's not started mixing placeholders with actual possible values please :)

addon/components/o-s-s/currency-input.hbs
```diff
@@ -1,22 +1,31 @@
...
+            <span>{{this.selectedCurrencySymbol}}</span>
+          {{/if}}
+          {{#if @onlyCurrency}}
+            <span class=" margin-right-px-12 {{if (eq this.selectedCurrencyCode "Select") "font-color-gray-500" }}">{{this.selectedCurrencyCode}}</span>
+          {{/if}}
+        </div>
+      </div>
```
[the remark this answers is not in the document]
phndiaye: +1 (and way easier to test)

addon/components/o-s-s/currency-input.ts
```diff
@@ -7,7 +7,8 @@ import { isEmpty } from '@ember/utils';
 interface OSSCurrencyInputArgs {
   currency: string;
   value: number;
-  onChange(currency: string, value: number): void;
+  onChange(currency: string | undefined, value: number): void;
```
phndiaye, line 10: I think it's the `value` that's supposed to be undefined in case of `onlyCurrency` here

addon/components/o-s-s/currency-input.ts
```diff
@@ -28,12 +29,18 @@ export default class OSSCurrencyInput extends Component<OSSCurrencyInputArgs> {
...
+    return this.selectedCurrency?.symbol || '—';
+  }
+
+  get selectedCurrencyCode(): string {
+    return this.selectedCurrency?.code || 'Select';
   }
 
   get selectedCurrency() {
-    if (isEmpty(this.args.currency)) {
+    if (isEmpty(this.args.currency) && this.args.onlyCurrency == false) {
```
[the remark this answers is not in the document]
sBenjamin270495: Thank you, i did that because the onlyCurrency can be undefined and  i thought that the typing of === could be a problem but not.
After some search on undifined status , i see that the problem is the logic of the function, i can simplify this function

addon/components/o-s-s/currency-input.hbs
```diff
@@ -1,21 +1,31 @@
...
-           autocomplete="off" />
+    {{#unless @onlyCurrency}}
+      <Input class="fx-1" type="number" @value={{this.localValue}} {{on "keydown" this.onlyNumeric}}
+             {{on "keyup" this.notifyChanges}}
+             autocomplete="off" />
+    {{/unless}}
   </div>
   {{#if this.currencySelectorShown}}
-    <OSS::InfiniteSelect @items={{this.filteredCurrencies}} @onSearch={{this.onSearch}}
+    <OSS::InfiniteSelect @items={{this.filteredCurrencies}} @onSearch={{this.onSearch}}
```
JulienVannier66, line 26:
```suggestion
    <OSS::InfiniteSelect @items={{this.filteredCurrencies}} @onSearch={{this.onSearch}}
```
extra space at the end

addon/components/o-s-s/currency-input.hbs
```diff
@@ -1,21 +1,31 @@
...
+             {{on "keyup" this.notifyChanges}}
+             autocomplete="off" />
+    {{/unless}}
   </div>
   {{#if this.currencySelectorShown}}
-    <OSS::InfiniteSelect @items={{this.filteredCurrencies}} @onSearch={{this.onSearch}}
+    <OSS::InfiniteSelect @items={{this.filteredCurrencies}} @onSearch={{this.onSearch}} 
                          @onSelect={{this.onSelect}}
-                         @searchPlaceholder={{t "oss-components.currency-input.search"}}
+                         @searchPlaceholder={{t "oss-components.currency-input.search" }}
```
JulienVannier66:
```suggestion
                         @searchPlaceholder={{t "oss-components.currency-input.search"}} 
```
extra space

addon/components/o-s-s/currency-input.ts
```diff
@@ -7,7 +7,8 @@ import { isEmpty } from '@ember/utils';
 interface OSSCurrencyInputArgs {
   currency: string;
   value: number;
-  onChange(currency: string, value: number): void;
+  onChange(currency: string , value: number ): void;
```
JulienVannier66, line 10:
```suggestion
  onChange(currency: string , value: number): void;
```

addon/components/o-s-s/currency-input.ts
```diff
@@ -31,14 +32,21 @@ export default class OSSCurrencyInput extends Component<OSSCurrencyInputArgs> {
...
+    if (this.emptyCurrency) {
       return this._currencies[0];
     }
-
     return this._currencies.find((currency: Currency) => currency.code === this.args.currency) || this._currencies[0];
   }
 
+  get emptyCurrency(): boolean {
+    return isEmpty(this.args.currency);
+  }
```
JulienVannier66, line 48: Is this getter used in the template?

app/styles/currency-input.less
```diff
@@ -51,4 +46,12 @@
...
     }
   }
+  .currency-input.onlycurrency {
+    padding-right: 0px;
+    .currency-selector {
+      border-right: none;
+      margin-right: 0px;
+      width: 100%;
+    }
+  }
```
JulienVannier66, line 56:
```suggestion
  .onlycurrency {
    padding-right: 0;
    .currency-selector {
      border-right: none;
      margin-right: 0;
      width: 100%;
    }
  }
```
Move this class in `currency-input` definition directly. You can delete `px` when you set 0. Do you need `width: 100%` here?

tests/integration/components/o-s-s/currency-input-test.ts
```diff
@@ -16,6 +16,49 @@ module('Integration | Component | o-s-s/currency-input', function (hooks) {
     assert.dom('.currency-input-container').exists();
   });
 
+  test('it renders currency only', async function (assert) {
+    this.onChange = () => {};
+    this.onlyCurrency = true;
+    await render(
+      hbs`<OSS::CurrencyInput @currency="USD" @onlyCurrency={{this.onlyCurrency}} @onChange={{this.onChange}} />`
```
JulienVannier66, line 23:
```suggestion
    await render(
      hbs`<OSS::CurrencyInput @currency="USD" @onlyCurrency={{true}} @onChange={{this.onChange}} />`
```
You don't use it in test, set up value directly in the component.

tests/integration/components/o-s-s/currency-input-test.ts
```diff
@@ -16,6 +16,49 @@ module('Integration | Component | o-s-s/currency-input', function (hooks) {
...
+    assert.dom('.currency-input input').doesNotExist();
+  });
+
+  test('it renders currency only with empty currency param', async function (assert) {
+    this.onChange = () => {};
+    this.onlyCurrency = true;
+    this.currency = null;
+    await render(
+      hbs`<OSS::CurrencyInput @currency={{this.currency}} @onlyCurrency={{this.onlyCurrency}} @onChange={{this.onChange}} />`
+    );
```
JulienVannier66, line 46:
```suggestion
    await render(
      hbs`<OSS::CurrencyInput @onlyCurrency={{true}} @onChange={{this.onChange}} />`
    );
```
You can simplify

tests/integration/components/o-s-s/currency-input-test.ts
```diff
@@ -54,14 +97,29 @@ module('Integration | Component | o-s-s/currency-input', function (hooks) {
...
       await click(clickableRows[4]);
       assert.dom('.currency-selector').hasText('A$');
     });
+    test('Selecting a new currency in the Currency selector triggers the onChange method with currency only', async function (assert) {
+      this.currency = '';
+      this.onChange = (currency: string, _: number) => {
+        assert.equal(currency, 'AUD');
+        this.set('currency', currency);
+      };
+      sinon.spy(this.onChange);
```
JulienVannier66, line 114: Use `sinon.stub` instead. You are not sure that the `assert.equal` is call here

addon/components/o-s-s/currency-input.hbs
```diff
@@ -1,21 +1,31 @@
...
         <i class="far fa-chevron-down margin-left-xxx-sm"></i>
       {{/if}}
     </div>
-    <Input class="fx-1" type="number" @value={{this.localValue}} {{on "keydown" this.onlyNumeric}}
-           {{on "keyup" this.notifyChanges}}
-           autocomplete="off" />
+    {{#unless @onlyCurrency}}
+      <Input class="fx-1" type="number" @value={{this.localValue}} {{on "keydown" this.onlyNumeric}}
+             {{on "keyup" this.notifyChanges}}
+             autocomplete="off" />
```
Miexil, line 22: There is space for the `autocomplete="off"` on the line above.

addon/components/o-s-s/currency-input.hbs
```diff
@@ -1,21 +1,31 @@
...
+    {{/unless}}
   </div>
   {{#if this.currencySelectorShown}}
-    <OSS::InfiniteSelect @items={{this.filteredCurrencies}} @onSearch={{this.onSearch}}
+    <OSS::InfiniteSelect @items={{this.filteredCurrencies}} @onSearch={{this.onSearch}} 
                          @onSelect={{this.onSelect}}
-                         @searchPlaceholder={{t "oss-components.currency-input.search"}}
+                         @searchPlaceholder={{t "oss-components.currency-input.search"}}
```
Miexil, line 28: extra space at the end

addon/components/o-s-s/currency-input.ts
Miexil: +1 
Unless used in the template, the `isEmtpy(this.args.currency)` should just be added b ack to line 40.

addon/components/o-s-s/currency-input.hbs
sBenjamin270495: ok but the indentation here and before I do changes are the same

tests/integration/components/o-s-s/currency-input-test.ts
```diff
@@ -16,6 +16,44 @@ module('Integration | Component | o-s-s/currency-input', function (hooks) {
...
+    assert.dom('.currency-input-container').exists();
+    assert.dom('.currency-selector').exists();
+
+    const currencyLabel = await findAll('.currency-selector span');
+
+    assert.dom(currencyLabel[0]).exists();
+    assert.dom(currencyLabel[0]).hasText('$');
+
+    assert.dom(currencyLabel[1]).exists();
+    assert.dom(currencyLabel[1]).hasText('USD');
```
phndiaye, line 52: maybe a `assert.dom('.currency-selector').hasText('$ USD')` can replace all this?  🤷‍♂️ (double check but it's good anyway)

tests/integration/components/o-s-s/currency-input-test.ts
```diff
@@ -16,6 +16,44 @@ module('Integration | Component | o-s-s/currency-input', function (hooks) {
     assert.dom('.currency-input-container').exists();
   });
 
+  test('it renders currency only', async function (assert) {
```
phndiaye, line 19: what about nesting all the tests you added into a `module('currency only mode')` ?

phndiaye (approved):

addon/components/o-s-s/currency-input.ts
```diff
@@ -7,7 +7,8 @@ import { isEmpty } from '@ember/utils';
 interface OSSCurrencyInputArgs {
   currency: string;
   value: number;
-  onChange(currency: string, value: number): void;
+  onChange(currency: string , value: number): void;
```
Miexil, line 10: Extra space.

JulienVannier66 (approved):

Miexil:
Screenshot 2022-07-12 at 10 18 36

Also, you need to add a space-between the chevron and the currency + label in order to align the chevron to the right of the input.

sBenjamin270495:
> Screenshot 2022-07-12 at 10 18 36
> 
> Also, you need to add a space-between the chevron and the currency + label in order to align the chevron to the right of the input.

There are already a space between on div .currency-selector with class fx-malign-space-between, the problem is when i remove the width : 100% i did bad test sorry...

Miexil (approved):

PR #200, opened 2022-08-02
Merge @upfluence/oss package into this project

phndiaye:
### What does this PR do?

For some time now, we have noticed that the split between this package (holding Ember specific code) and @upfluence/oss package (holding more agnostic, Less CSS + assets only code) is not viable anymore and brings some pain point that are a bit annoying:
- having to open PRs in two projects for the same task
- Yarn links w/ @upfluence/oss that are not always straighforward (even more if you link this package to another running app)
- Documentation being split between two packages

After discussion w/ the team, this PR brings the CSS, assets and deps from the @upfluence/oss package into @upfluence/oss-components definitely.

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled


app/styles/base/_all.less
```diff
@@ -0,0 +1,23 @@
...
+@import '_inputs';
+@import '_inputs-range';
+@import '_links';
+@import '_loading';
+@import '_navs';
+@import '_notifications';
+@import '_progress';
+@import '_slider';
+@import '_stats';
+@import '_tables'; // TO_BE_DEPRECIATED
```
Miexil, line 17: deprecated* 😇

Miexil (approved):
Now, let's see the damage on staging :D

PR #201, opened 2022-08-04
Power select as dropdown

phndiaye:
### What does this PR do?

- Use PowerSelect as dropdown instead of sticky infinite select
- Updated the Chip to remove its margins as it's not its responsibility to set it 
- Update the ArrayInput component to add said margins

Screenshot 2022-08-04 at 16 11 18

Screenshot 2022-08-04 at 16 15 47

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/array-input.hbs
+++ b/addon/components/o-s-s/array-input.hbs
@@ -1,4 +1,4 @@
-<div class="array-input-container fx-row fx-xalign-center" ...attributes>
+<div class="array-input-container fx-row fx-xalign-center fx-gap-px-6 padding-left-px-6 padding-right-px-6" ...attributes>
   {{#each this.items as |item index|}}
     <OSS::Chip @label={{item}} @onRemove={{fn this.removeItem index}} />
   {{/each}}
--- a/addon/components/o-s-s/chip.hbs
+++ b/addon/components/o-s-s/chip.hbs
@@ -1,4 +1,4 @@
-<div class={{this.computedClass}} ...attributes>
+<div class={{this.computedClass}} ...attributes {{on "click" this.stopPropagation}}>
   <span class="text-style-semibold {{if @maxDisplayWidth "chip-ellipsis"}}"
         style={{if @maxDisplayWidth this.ellipsisStyle}}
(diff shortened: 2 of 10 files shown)


addon/components/o-s-s/chip.hbs
```diff
@@ -1,4 +1,4 @@
-<div class={{this.computedClass}} ...attributes>
+<div class={{this.computedClass}} ...attributes {{on "click" this.stopPropagation}}>
```
Miexil, line 1: hmmmm ? 🤔 
Why ?

addon/components/o-s-s/array-input.hbs
```diff
@@ -1,4 +1,4 @@
-<div class="array-input-container fx-row fx-xalign-center" ...attributes>
+<div class="array-input-container fx-row fx-xalign-center fx-gap-px-6 padding-left-px-6 padding-right-px-6" ...attributes>
```
Miexil, line 1: 80% chance where we use array-input will look weird after we merge this in.
I'm pretty sure I adjusted some css where we used some of these components... Nothing too awful though, we can fix them when we see them 🤷

addon/components/o-s-s/power-select.hbs
```diff
@@ -1,24 +1,33 @@
-<div class="upf-power-select fx-col" ...attributes
+<div class="upf-power-select fx-1 fx-col" ...attributes
```
Miexil, line 1: 🙏 thanks.

addon/components/o-s-s/array-input.hbs
phndiaye: I double checked in the CCs and BCCs in the composer and it works well w/ the adjustments done on the Chip too.

addon/components/o-s-s/chip.hbs
phndiaye: Because when used in the PowerSelect it triggers the click outside + the only thing we want to monitor clicks on here is the cross icon

Miexil (approved):

PR #207, opened 2022-09-21
Add extra-content named block for OSS::Alert

JulienVannier66:
### What does this PR do?

Add named block `extra-content` for OSS::Alery

Related to: https://github.com/upfluence/backlog/issues/1811

Screenshot from 2022-09-21 18-27-21

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/alert.hbs
+++ b/addon/components/o-s-s/alert.hbs
@@ -11,6 +11,9 @@
       {{#if @subtitle}}
         <span class="subtitle">{{@subtitle}}</span>
       {{/if}}
+      {{#if (has-block "extra-content")}}
+        {{yield to="extra-content"}}
+      {{/if}}
     </div>
   </div>
 </div>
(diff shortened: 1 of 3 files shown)


addon/components/o-s-s/alert.hbs
```diff
@@ -11,6 +11,9 @@
       {{#if @subtitle}}
         <span class="subtitle">{{@subtitle}}</span>
       {{/if}}
+      {{#if (has-block "extra-content")}}
```
Miexil, line 14: The "main-container" class is defined as fx-row ; you might want to add an fx-col wrapping everything so that we don't get bad surprises when using the extra content !

phndiaye: @Miexil it's in the `text-container` here. which is defined as `col`

phndiaye (approved):

sBenjamin270495 (approved):

addon/components/o-s-s/alert.hbs
JulienVannier66: Do you want I move the block in `main-container` and did the update or keep it in `text-container` @Miexil ?

Miexil (approved):

PR #214, opened 2022-09-29
Upload area component

phndiaye:
### What does this PR do?

Add support for new OSS::UploadArea component with:
- disabled mode support
- drag and drop unless disabled
- manual file section via the OS file browser
- file validation (via both drag & drop and manual selection)

Related to: https://github.com/upfluence/backlog/issues/2004

Screenshot 2022-09-29 at 10 22 18

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled


tests/integration/components/o-s-s/upload-area-test.ts
```diff
@@ -0,0 +1,204 @@
...
+
+        assert.ok(
+          toastStub.calledWith(
+            intlService.t(`oss-components.upload-area.errors.filesize.description`, { max_filesize: '1B' }),
+            intlService.t(`oss-components.upload-area.errors.filesize.title`)
+          )
+        );
+      });
+
+      //test('the file display state is entered if the dropped file passes the validation', async function (assert) {});
```
phndiaye, line 201: this will be written the moment i have the UploadItem component

addon/components/o-s-s/upload-area.hbs
```diff
@@ -0,0 +1,42 @@
+{{#if this.hasFile}}
+{{else}}
```
JulienVannier66, line 2:
```suggestion
{{#unless this.hasFile}}
```

addon/components/o-s-s/upload-area.ts
```diff
@@ -0,0 +1,151 @@
...
+  @action
+  _onDragLeave(): void {
+    if (this.args.disabled) return;
+    this.dragging = false;
+  }
+
+  @action
+  _onDrop(event: DragEvent): void {
+    event.stopPropagation();
+    event.preventDefault();
```
JulienVannier66, line 106: Need the 2 functions?

addon/components/o-s-s/upload-area.ts
```diff
@@ -0,0 +1,151 @@
...
+      this._handleFileUpload(file);
+    }
+
+    this.dragging = false;
+  }
+
+  private _handleFileUpload(file: File): void {
+    if (this._validateFile(file)) {
+      this.selectedFile = file;
+      return;
```
JulienVannier66, line 119: Necessary?

translations/en-us.yaml
```diff
@@ -31,3 +31,15 @@ oss-components:
   select:
     search: Search
     placeholder: Select
+  upload-area:
+    drop_file:
+      default: Drop your file here, or
+      disabled: Drop your file here, or browse
+    browse: "browse"
```
JulienVannier66, line 38: Do you need `"` for browse key

translations/en-us.yaml
```diff
@@ -31,3 +31,15 @@ oss-components:
...
+    errors:
+      filetype:
+        title: Upload error
+        description: The type of the file you uploaded is not allowed.
+      filesize:
+        title: Upload error
+        description: "The file you uploaded is too heavy. Maximum file size: {max_filesize}"
```
JulienVannier66, line 45: Same here?

addon/components/o-s-s/upload-area.hbs
phndiaye: @JulienVannier66 That's where i plan to display the uploaded file, that's why i let the `if` even if empty

translations/en-us.yaml
phndiaye: I tend to quote them when i use a variable in the trad but it's not mandatory

addon/components/o-s-s/upload-area.ts
phndiaye: - stopPropagation to avoid side effects
- preventDefault is to avoid browsers from opening the file in a tab (like Chrome does for example)

Put both to be extra sure, but the preventDefault only works too

JulienVannier66 (approved):

Miexil (approved):

PR #257, opened 2023-01-13
currency-input: add support placeholder and for disabling the currency value update

phndiaye:
### What does this PR do?

currency-input: add support placeholder and for disabling the currency value update

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/currency-input.ts
+++ b/addon/components/o-s-s/currency-input.ts
@@ -36,7 +36,7 @@ export default class OSSCurrencyInput extends Component<OSSCurrencyInputArgs> {
   }
 
   get allowCurrencyUpdate(): boolean {
-    return typeof this.args.allowCurrencyUpdate === 'boolean' ? this.args.allowCurrencyUpdate : true;
+    return this.args.allowCurrencyUpdate ?? true;
   }
 
   get selectedCurrencySymbol(): string {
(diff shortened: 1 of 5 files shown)


addon/components/o-s-s/currency-input.ts
```diff
@@ -18,17 +20,25 @@ export default class OSSCurrencyInput extends Component<OSSCurrencyInputArgs> {
...
-  @tracked localValue: number = this.args.value || 0;
+  @tracked localValue: number = this.args.value;
 
   constructor(owner: unknown, args: OSSCurrencyInputArgs) {
     super(owner, args);
 
+    if (!this.args.value && !this.args.placeholder) {
+      this.localValue = 0;
+    }
+
```
phndiaye, line 31: this is to avoid breaking current usages of the component. this way, we skip the default value only if there is a placeholder provided

Miexil: ^^ okay !

Miexil (approved):

tests/integration/components/o-s-s/currency-input-test.ts
```diff
@@ -101,6 +101,15 @@ module('Integration | Component | o-s-s/currency-input', function (hooks) {
       this.set('currency', 'EUR');
       assert.dom('.currency-selector').hasText('€');
     });
+    test('the selection is disabled if @allowCurrencyUpdate is falsy', async function (assert) {
```
JulienVannier66, line 104: Can you add a test for `placeholder` attribute also?

JulienVannier66 (approved):
LGTM

PR #258, opened 2023-01-17
Alias bold to semibold (and change wherever possible right now)

phndiaye:
### What does this PR do?

Alias bold to semibold (and change wherever possible right now)

Related to: https://github.com/upfluence/backlog/issues/1807

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/app/styles/core-v2/_shims.less
+++ b/app/styles/core-v2/_shims.less
@@ -1,16 +1,14 @@
 /* cyrillic-ext */
 @font-face {
   font-family: 'Inter';
-  font-style: normal;
   font-weight: 700;
   font-display: swap;
-  src: url("https://github.com/rsms/inter/raw/master/docs/font-files/Inter-SemiBold.woff2") format('woff2');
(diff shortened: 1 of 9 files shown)


app/styles/core-v2/_shims.less
```diff
@@ -0,0 +1,63 @@
+/* cyrillic-ext */
```
JulienVannier66, line 1: What is the goal of this file?

phndiaye: It forces Bold/700 font-weights to load the SemiBold font. This way we don't have to chase all Bold usages right now.

JulienVannier66: So it will be impossible to set bold in the soft after that

phndiaye: Yep !

Miexil (approved):

JulienVannier66 (approved):

PR #264, opened 2023-02-21
components/oss/array-input: add support for extra keyboard triggers

phndiaye:
### What does this PR do?

Allow the `OSS::ArrayInput` component to receive extra keyboard event codes that can be used to validate an entry beside the default Enter one.

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/array-input.ts
+++ b/addon/components/o-s-s/array-input.ts
@@ -10,6 +10,8 @@ interface OSSArrayInputArgs {
   placeholder?: string;
 }
 
+const DEFAULT_KEYBOARD_TRIGGERS = ['Enter'];
+
 export default class OSSArrayInput extends Component<OSSArrayInputArgs> {
   @tracked currentValue = '';
   @tracked items: string[] = [];
@@ -22,7 +24,7 @@ export default class OSSArrayInput extends Component<OSSArrayInputArgs> {
   }
 
   get keyboardTriggers(): string[] {
-    return ['Enter'].concat(this.args.keyboardTriggers || []);
+    return DEFAULT_KEYBOARD_TRIGGERS.concat(this.args.keyboardTriggers || []);
   }
 
   private _triggerComponentRedraw(): void {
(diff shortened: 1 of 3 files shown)


addon/components/o-s-s/array-input.ts
```diff
@@ -20,6 +21,10 @@ export default class OSSArrayInput extends Component<OSSArrayInputArgs> {
     }
   }
 
+  get keyboardTriggers(): string[] {
+    return ['Enter'].concat(this.args.keyboardTriggers || []);
```
phndiaye, line 25: Move Enter to const

JulienVannier66 (approved):
After set `Enter` as default trigger, LGTM

addon/components/o-s-s/array-input.ts
```diff
@@ -59,7 +66,8 @@ export default class OSSArrayInput extends Component<OSSArrayInputArgs> {
 
   @action
   keyListener(keyboardEvent: KeyboardEvent): void {
-    if (keyboardEvent.code === 'Enter' && this.currentValue.length > 0) {
+    if (this.keyboardTriggers.includes(keyboardEvent.code) && this.currentValue.length > 0) {
```
Miexil, line 69: are you sure you want to check against codes rather than keys ?

phndiaye: the arg is like `['Space']`, I thought it was better like that and more "human"

Miexil: After discussing live, we chose to use the `code` instead of `key`
For reference, from the MDN :

> For example, the code returned is “KeyQ” for the Q key on a QWERTY layout keyboard, but the same code value also represents the ' key on Dvorak keyboards and the A key on AZERTY keyboards. That makes it impossible to use the value of code to determine what the name of the key is to users if they’re not using an anticipated keyboard layout.
To determine what character corresponds with the key event, use the KeyboardEvent.key property instead.

Miexil (approved):

PR #267, opened 2023-03-24
OSS::Button redesign

phndiaye:
### What does this PR do?

Mainly implements the new design for the buttons as in : https://www.figma.com/file/N7SDuH9mvC9zTlsLN49N4L/2.0-UI-Kit-%E2%9D%96?node-id=1102-266&t=VQYnEgqG3BDp9mD9-0.

This PR also slightly updates the values of the existing color palette and introduces a new "Melon" color to the extended palette.

It also removes the legacy LoadingButton & DestructiveButton components.

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled


app/styles/atoms/button.less
```diff
@@ -1,22 +1,35 @@
...
-  transition: all 0.2s ease;
+
+  i,
+  span {
+    color: var(~'--color-@{variant}-500');
+  }
+
+  a& {
+    line-height: 2.4rem;
+  }
```
phndiaye, line 13: unfortunately we have to keep these as there are still some places where we used buttons styles on `a` tags (on some `LinkTo` for example)

Miexil (approved):

PR #269, opened 2023-04-07
[Tertiary btn] Remove padding for tertiary button

JulienVannier66:
### What does this PR do?

Remove padding for tertiary skin button

Related to: #

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/app/styles/atoms/button.less
+++ b/app/styles/atoms/button.less
@@ -59,7 +59,8 @@
     .text-size-7;
   }
 
-  i, span {
+  i,
+  span {
     transition: all 0.25s ease-in-out;
   }
 
@@ -91,7 +92,8 @@
 
   transition: all 0.25s ease-in-out;
 
-  i, span {
+  i,
+  span {
     transition: all 0.25s ease-in-out;
   }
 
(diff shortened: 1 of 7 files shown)


app/styles/atoms/button.less
```diff
@@ -91,7 +92,8 @@
 
   transition: all 0.25s ease-in-out;
 
-  i, span {
+  i,
+  span {
```
Miexil, line 96: Is this prettier's doing again ? 😮‍💨

Miexil (commented):
I think you already explained last friday but I forgot.
What's the point of doing this ?
We already have an OSS::Link component

app/styles/atoms/button.less
JulienVannier66: Yes!

JulienVannier66:
The goal is to have the same design for OSS::Button tertiary and OSS::Link, it is not the case currently. It is help us for future change on it. The usage between these two elements is tech only, maybe we can deprecate one of these in the future without layout update

phndiaye (approved):

phndiaye (approved):

Miexil (approved):

PR #291, opened 2023-07-05
Add: OSS::Skeleton component

aprentout:
### What does this PR do?
Add new component OSS::Skeleton

Related to: [#ENG-510](https://linear.app/upfluence/issue/ENG-510/ossskeleton)

Capture d’écran 2023-07-05 à 16 34 01

https://github.com/upfluence/oss-components/assets/111493996/6ec8cf16-e6f4-4c65-9432-a3cbada69469

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/skeleton.stories.js
+++ b/addon/components/o-s-s/skeleton.stories.js
@@ -12,7 +12,7 @@ export default {
         type: {
           summary: 'number'
         },
-        defaultValue: { summary: '100' }
+        defaultValue: { summary: '36' }
       },
       control: { type: 'number' }
     },
@@ -22,12 +22,12 @@ export default {
         type: {
           summary: 'number'
         },
-        defaultValue: { summary: '100' }
+        defaultValue: { summary: '36' }
       },
       control: { type: 'number' }
     },
     multiple: {
-      description: 'How many skeleton effect should be displayed',
+      description: 'How many skeleton effects should be displayed',
       table: {
         type: {
           summary: 'number'
@@ -46,6 +46,16 @@ export default {
       },
       control: { type: 'number' }
     },
+    randomize: {
+      description: 'Randomize skeleton effect width',
+      table: {
+        type: {
+          summary: 'boolean'
+        },
+        defaultValue: { summary: false }
+      },
+      control: { type: 'boolean' }
+    },
     type: {
       description: 'Direction of the skeleton',
       table: {
@@ -61,7 +71,7 @@ export default {
   parameters: {
     docs: {
       description: {
-        component: 'Component used to create skeleton effect.'
+        component: 'Component used to create a skeleton effect.'
       },
       iframeHeight: 250
     }
@@ -72,13 +82,16 @@ const defaultArgs = {
   height: 200,
   width: 300,
   multiple: 1,
-  gap: 9
+  gap: 9,
+  randomize: false
 };
 
 const Template = (args) => ({
   template: hbs`
     <div class="bg-color-white padding-px-6">
-      <OSS::Skeleton @height={{this.height}} @width={{this.width}} @multiple={{this.multiple}} @gap={{this.gap}} @type={{this.type}}/>
+      <OSS::Skeleton @height={{this.height}} @width={{this.width}} @multiple={{this.multiple}} @gap={{this.gap}} 
+                     @type={{this.type}}
+                     @randomize={{this.randomize}}/>
     </div>
   `,
   context: args
--- a/addon/components/o-s-s/skeleton.stories.js
+++ b/addon/components/o-s-s/skeleton.stories.js
@@ -1,0 +1,88 @@
+import hbs from 'htmlbars-inline-precompile';
+
+const DirectionTypes = ['row', 'column'];
+
+export default {
+  title: 'Components/OSS::Skeleton',
+  component: 'skeleton',
+  argTypes: {
+    height: {
+      description: 'Box height in px',
+      table: {
+        type: {
+          summary: 'number'
+        },
+        defaultValue: { summary: '100' }
+      },
+      control: { type: 'number' }
+    },
+    width: {
+      description: 'Box width in px',
+      table: {
+        type: {
+          summary: 'number'
+        },
+        defaultValue: { summary: '100' }
+      },
+      control: { type: 'number' }
+    },
+    multiple: {
+      description: 'How many skeleton effect should be displayed',
+      table: {
+        type: {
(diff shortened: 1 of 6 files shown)


addon/components/o-s-s/skeleton.ts
```diff
@@ -0,0 +1,83 @@
...
+interface Row {
+  style?: string;
+}
+
+interface OSSSkeletonArgs {
+  width: number;
+  height: number;
+  multiple?: number;
+  type?: 'row' | 'column';
+  gap?: number;
```
Miexil, line 13: Can we also put the randomization as a parameter please ? 😇

addon/components/o-s-s/skeleton.ts
```diff
@@ -0,0 +1,83 @@
...
+export default class OSSSkeleton extends Component<OSSSkeletonArgs> {
+  constructor(owner: unknown, args: OSSSkeletonArgs) {
+    super(owner, args);
+
+    if (this.args.type) {
+      assert(
+        `[component][OSS::Skeleton] The @type argument should be a value of ${['row', 'column']}`,
+        ['row', 'column'].includes(this.args.type)
+      );
+    }
```
Miexil, line 25: maybe assert width & height as well as they are not optional values ?

Miexil: Ah, I see there are default values below, maybe tag them as optional in the OSSSkeletonArgs interface then ?

addon/components/o-s-s/skeleton.ts
```diff
@@ -0,0 +1,83 @@
...
+    } else {
+      for (let i = 0; i < multiple; i++) {
+        rows.push(this._generateRandomRowSize());
+      }
+    }
+
+    return rows;
+  }
+
+  _generateRandomRowSize(): Row {
```
Miexil, line 74: remove `_` add `private` keyword instead

addon/components/o-s-s/skeleton.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (this.args.type) {
+      assert(
+        `[component][OSS::Skeleton] The @type argument should be a value of ${['row', 'column']}`,
+        ['row', 'column'].includes(this.args.type)
+      );
+    }
+  }
+
+  get height(): number {
+    return this.args.height || 100;
```
Miexil, line 29: default height should be something more reasonable, maybe `36px` to match  our input height ?

tests/integration/components/o-s-s/skeleton-test.ts
```diff
@@ -0,0 +1,149 @@
...
+  });
+
+  module('@height parameters', () => {
+    test('Default height', async function (assert) {
+      await render(hbs`<OSS::Skeleton/>`);
+
+      assert.dom('.upf-skeleton-content').hasStyle({ height: '100px' });
+    });
+
+    test('Then style height should correspond to parameter value', async function (assert) {
```
Miexil, line 21: The* ?

tests/integration/components/o-s-s/skeleton-test.ts
```diff
@@ -0,0 +1,149 @@
...
+  });
+
+  module('@width parameters', () => {
+    test('Default width', async function (assert) {
+      await render(hbs`<OSS::Skeleton/>`);
+
+      assert.dom('.upf-skeleton-content').hasStyle({ width: '100px' });
+    });
+
+    test('Then style width should correspond to parameter value', async function (assert) {
```
Miexil, line 37: The* ?

tests/integration/components/o-s-s/skeleton-test.ts
```diff
@@ -0,0 +1,149 @@
...
+  module('@multiple parameters', () => {
+    test('Default has one skeleton effect', async function (assert) {
+      await render(hbs`<OSS::Skeleton/>`);
+
+      let items = findAll('.upf-skeleton-content .upf-skeleton-effect');
+
+      assert.ok(items.length === 1);
+    });
+
+    test('Then has multiple skeleton effect', async function (assert) {
```
Miexil, line 71: what does `then` refer to here ? 🤔

tests/integration/components/o-s-s/skeleton-test.ts
```diff
@@ -0,0 +1,149 @@
...
+    });
+  });
+
+  module('When type is row', () => {
+    hooks.beforeEach(function () {
+      this.multiple = 4;
+      this.width = 200;
+    });
+
+    test('Then content should have class fx-col', async function (assert) {
```
Miexil, line 88: Ah, I think I understand the `then`s I've commented above.
`Then` makes sense if there is a `when` in the module. here you should add `Then the content`

addon/components/o-s-s/skeleton.stories.js
```diff
@@ -0,0 +1,88 @@
...
+      table: {
+        type: {
+          summary: 'number'
+        },
+        defaultValue: { summary: '100' }
+      },
+      control: { type: 'number' }
+    },
+    multiple: {
+      description: 'How many skeleton effect should be displayed',
```
Miexil, line 30: effects*
or rows

addon/components/o-s-s/skeleton.stories.js
```diff
@@ -0,0 +1,88 @@
...
+        defaultValue: { summary: 'row' }
+      },
+      options: DirectionTypes,
+      control: { type: 'select' }
+    }
+  },
+  parameters: {
+    docs: {
+      description: {
+        component: 'Component used to create skeleton effect.'
```
Miexil, line 64: to create a* skeleton effect

Miexil (commented):
BTW, looking at your screenshot, we will never use bars going down, what I meant with the `row` and `column` issue was more like a flex-direction.
row -> puts the skeleton bars next to each other on the same horizontal line
column -> puts the skeleton bars underneath each other on a vertical line

addon/components/o-s-s/skeleton.ts
JulienVannier66: So here `type` is also mandatory? It is not the case in the interface

aprentout: `if (this.args.type)` prevent mandatory

Miexil (approved):

JulienVannier66 (approved):

phndiaye (approved):

PR #314, opened 2023-10-17
Added typecheck for IconNames

OwenCoogan:
### What does this PR do?

It adds typecheck for icon name to avoid flailing around looking for IconNames
Related to: #
https://linear.app/upfluence/issue/ENG-1138/type-icon-properties-on-ossicon-component

Just a new iconNames file & added typecheck on OSS::Icon component

### Good PR checklist
- [x] Title makes sense
- [ ] Properly assigned
- [ ] Added/updated tests
- [ ] Properly labeled


package.json
```diff
@@ -55,6 +55,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.11.6",
+    "els-addon-typed-templates": "^4.2.4",
```
JulienVannier66, line 58: Remove unused lib

addon/components/o-s-s/iconName.enum.ts
JulienVannier66, line 1: Can you order it alphabetically?

addon/components/o-s-s/icon.ts
```diff
@@ -1,10 +1,11 @@
 import Component from '@glimmer/component';
 import { assert } from '@ember/debug';
+import { IconNames } from './iconName.enum';
 
 export type IconStyle = 'solid' | 'regular' | 'light' | 'duotone' | 'brand';
 
 interface OSSIconArgs {
-  icon: string;
+  icon: IconNames;
```
Miexil, line 8: Just typing it as a message so that other reviewers don't ask you.

It looks to me like you are missing changes in the icon.hbs file.

Currently if I use something like :
`<OSS::Icon @icon="Analytics" />`

The template will be rendered as :
`<i class="far Analytics"></i>` (with `far` being the default style value)

We need a way to convert "Analytics" back to "fa-analytics" so that font-awesome can match to its own name.

Miexil: UNLESS
You were thinking of using something to make that enum match directly the proper value in the template.
But then we'll need some other tool/library to handle this.

🤷

JulienVannier66 (approved):

Miexil (approved):

PR #319, opened 2023-10-25
Updated OSS::Copy component

OwenCoogan:
### What does this PR do?
It adds the inline attribute to copy component & related changes.

Related to: #
https://linear.app/upfluence/issue/ENG-1556/[oss-components]-osscopy-updated

https://www.loom.com/share/ffd978744530485dbec2a7759cdb1e48?sid=99516a38-df65-4110-96d9-f973d4524282

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/copy.hbs
+++ b/addon/components/o-s-s/copy.hbs
@@ -1,3 +1,7 @@
-<OSS::Button class={{if this.inline 'oss-copy--inline'}}
-             @icon="far fa-copy" @square={{true}} @size="sm" {{on "click" this.copy}}
+{{#if this.inline}}
+  <OSS::Icon @icon="fa-copy" class="oss-copy--inline" {{on "click" this.copy}}
              {{enable-tooltip placement="top" title=(t "oss-components.copy.tooltip") trigger="hover"}} />
+{{else}}
+  <OSS::Button @icon="far fa-copy" @square={{true}} @size="sm" {{on "click" this.copy}}
+               {{enable-tooltip placement="top" title=(t "oss-components.copy.tooltip") trigger="hover"}} />
+{{/if }}
--- a/addon/components/o-s-s/copy.hbs
+++ b/addon/components/o-s-s/copy.hbs
@@ -1,2 +1,3 @@
-<OSS::Button @icon="far fa-copy" @square={{true}} @size="sm" {{on "click" this.copy}}
+<OSS::Button class={{if this.inline 'oss-copy--inline'}}
+             @icon="far fa-copy" @square={{true}} @size="sm" {{on "click" this.copy}}
              {{enable-tooltip placement="top" title=(t "oss-components.copy.tooltip") trigger="hover"}} />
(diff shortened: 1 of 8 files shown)


addon/components/o-s-s/copy.ts
```diff
@@ -1,16 +1,19 @@
...
 
 interface OSSCopyArgs {
   value: string;
+  inline?: boolean;
 }
 
 export default class OSSCopy extends Component<OSSCopyArgs> {
   @service intl: any;
   @service declare toast: ToastService;
+  @tracked inline: boolean = this.args.inline || false;
```
JulienVannier66, line 16: Nit: use `??` instead of `||` because you want to set `false` for undefined value only

app/styles/atoms/copy.less
```diff
@@ -0,0 +1,11 @@
+.oss-copy--inline {
+  border: 0;
+  background-color: transparent;
+  color: var(--color-gray-500);
+  &:hover {
```
JulienVannier66, line 5: Add space between this two lines

JulienVannier66 (approved):
Otherwise LGTM

addon/components/o-s-s/copy.hbs
```diff
@@ -1,2 +1,3 @@
-<OSS::Button @icon="far fa-copy" @square={{true}} @size="sm" {{on "click" this.copy}}
+<OSS::Button class={{if this.inline 'oss-copy--inline'}}
+             @icon="far fa-copy" @square={{true}} @size="sm" {{on "click" this.copy}}
              {{enable-tooltip placement="top" title=(t "oss-components.copy.tooltip") trigger="hover"}} />
```
phndiaye, line 3: IMO it would be better to have a whole if/else here and not override buttons style. And just display an raw icon in a span/div

addon/components/o-s-s/copy.hbs
```diff
@@ -1,2 +1,8 @@
-<OSS::Button @icon="far fa-copy" @square={{true}} @size="sm" {{on "click" this.copy}}
+{{#if this.inline}}
+  <OSS::Icon @icon="fa-copy" class="oss-copy--inline"
              {{enable-tooltip placement="top" title=(t "oss-components.copy.tooltip") trigger="hover"}} />
+{{else}}
+  <OSS::Button class={{if this.inline 'oss-copy--inline'}}
```
aprentout, line 5: You could probably remove the `class` as there is an if before

JulienVannier66, line 5:
```suggestion
  <OSS::Button
```
You don't need this condition now

aprentout (approved):
Otherwise lgtm

addon/components/o-s-s/copy.hbs
```diff
@@ -1,2 +1,8 @@
-<OSS::Button @icon="far fa-copy" @square={{true}} @size="sm" {{on "click" this.copy}}
+{{#if this.inline}}
+  <OSS::Icon @icon="fa-copy" class="oss-copy--inline"
```
JulienVannier66, line 2: I don't on click action on it, it is normal?

JulienVannier66 (approved):

PR #320, opened 2023-10-26
Added OSS::Attribute::Text component

OwenCoogan:
### What does this PR do?
Added OSS::Attribute::Text to oss-components , a new component to display a string with label & copy that string to clipboard.

Related to: #
https://linear.app/upfluence/issue/ENG-1557/[oss-component]-new-ossattributetext-component

https://www.loom.com/share/3a5a070a58c74ab7a344831cd8086400?sid=d87c04ec-a9f0-4ac2-9f54-9bb54df5ade0

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled


app/styles/atoms/attribute-text.less
Miexil, line 1: I shared with you  a common `attribute` via slack, the goal was to have a common set of attribute classes that can be shared by the multiple types (text, country, etc.).
Can you use this one as base and what you need ?

addon/components/o-s-s/attribute-text.ts
```diff
@@ -0,0 +1,32 @@
...
+  get label(): string {
+    assert(`[component][OSS::Attribute::Text] label is required `, this.args.label);
+    return this.args.label;
+  }
+
+  get value(): string | undefined {
+    return this.args.value === undefined || this.args.value.length === 0 ? '-' : this.args.value;
+  }
+
+  get copyable(): boolean {
```
Miexil, line 29: @JulienVannier66 wasn't the copy button supposed to be displayed on hover of the componentrow ? 🤔

addon/components/o-s-s/attribute-text.stories.js
```diff
@@ -0,0 +1,73 @@
...
+        },
+        defaultValue: { summary: false }
+      },
+      control: { type: 'boolean' }
+    }
+  },
+  parameters: {
+    docs: {
+      description: {
+        component: 'A text component designed to have a label , a value & a copy value button'
```
Miexil, line 55: nit: no space before the comma

addon/components/o-s-s/attribute-text.hbs
```diff
@@ -0,0 +1,18 @@
+<div class="oss-attribute--text">
+  <div class="oss-attribute--text--label">
+    {{this.label}}
+    {{#if this.args.tooltip}}
+      <OSS::Icon @icon="far fa-info-circle" class="oss-attribute--info" @inline={{true}}
+                 {{enable-tooltip title=this.args.tooltip placement="top"}} />
+    {{/if}}
+
```
Miexil, line 8: random line break

app/styles/atoms/attribute-text.less
Miexil: Also your file is not linted

tests/integration/components/o-s-s/attribute-text-test.ts
```diff
@@ -0,0 +1,53 @@
...
+      assert.dom('.oss-attribute--info').exists();
+    });
+
+    test('it renders a dash as value when no value is specified', async function (assert) {
+      await render(hbs`<OSS::Attribute::Text @label="Hello" />`);
+      assert.dom('.oss-attribute--text').exists();
+      assert.dom('.oss-attribute--text--value').hasText('-');
+    });
+
+    test('it copies the value', async function (assert) {
```
Miexil, line 33: Missing a couple of tests:
`The text is copyable by default`
`The text is copyable if the @copyable argument is truthy`
`The text is not copyable (no copy button) if the argument is falsy`

addon/components/o-s-s/attribute-text.ts
Miexil: Also copyable is supposed to be truthy by default, pretty sure your condition below does not match the requirements

addon/components/o-s-s/attribute-text.hbs
```diff
@@ -0,0 +1,18 @@
+<div class="oss-attribute--text">
+  <div class="oss-attribute--text--label">
+    {{this.label}}
```
JulienVannier66, line 3: Put the label in a `span`

addon/components/o-s-s/attribute-text.ts
```diff
@@ -0,0 +1,32 @@
...
+export default class OSSAttributeText extends Component<OSSAttributeTextArgs> {
+  @tracked hasError: boolean = false;
+  @tracked displayInitials: boolean = false;
+
+  constructor(owner: unknown, args: OSSAttributeTextArgs) {
+    super(owner, args);
+  }
+
+  get label(): string {
+    assert(`[component][OSS::Attribute::Text] label is required `, this.args.label);
```
JulienVannier66, line 21: Set the `assert` in the constructor. You can remove this getter and use `@label` in the template directly

addon/components/o-s-s/attribute-text.ts
```diff
@@ -0,0 +1,32 @@
...
+    return this.args.label;
+  }
+
+  get value(): string | undefined {
+    return this.args.value === undefined || this.args.value.length === 0 ? '-' : this.args.value;
+  }
```
JulienVannier66, line 27:
```suggestion
  get value(): string {
    return isBlank(this.args.value) ? '-' : this.args.value;
  }
```
Just check of the value is blank, otherwise return dash

addon/components/o-s-s/attribute-text.ts
JulienVannier66: @Miexil Yes it is. But in some case, for the gender for example, when hover the line, the OSS::Copy should stay hidden

addon/components/o-s-s/attribute-text.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  get copyable(): boolean {
+    return !!(this.args.copyable && this.args.value && this.args.value.length > 0);
```
JulienVannier66, line 30:
```suggestion
    return Bolean(this.args.copyable) && !isBlank(this.args.value)
```
Should do the trick (be sure that the default value for `this.args.copyable` is `true`)

app/styles/atoms/attribute-text.less
JulienVannier66, line 1: Your class nomenclature is a bit weird, what did you base it on (there are a lot of double hyphens)? We can sync of you want

Also, add space in your code, for example after block `&--label`, `&--text`, etc
Where do to take the css value? There is a lot of `px` value, use upfluence variable instead

app/styles/atoms/attribute-text.less
```diff
@@ -0,0 +1,34 @@
...
+    flex-direction: row;
+    align-items: center;
+    justify-content: start;
+    gap: 6px;
+    &--info {
+      align-items: center;
+    }
+  }
+  &--text {
+    color: var(--color-gray-900);
```
JulienVannier66, line 26: You don't use this class int the template, can you remove it

tests/integration/components/o-s-s/attribute-text-test.ts
```diff
@@ -0,0 +1,53 @@
...
+  module('Default behavior', function () {
+    test('it renders', async function (assert) {
+      await render(hbs`<OSS::Attribute::Text @label="Hello" @value="World" />`);
+      assert.dom('.oss-attribute--text').exists();
+    });
+
+    test('it renders without a value', async function (assert) {
+      await render(hbs`<OSS::Attribute::Text @label="Hello" />`);
+      assert.dom('.oss-attribute--text').exists();
+    });
```
JulienVannier66, line 20: The `assert` is the same, why writing 2 different tests? Test the real value rendered

tests/integration/components/o-s-s/attribute-text-test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    test('it renders with a tooltip when specified', async function (assert) {
+      await render(hbs`<OSS::Attribute::Text @label="Hello" @tooltip="Tooltip" />`);
+      assert.dom('.oss-attribute--info').exists();
+    });
```
JulienVannier66, line 25: Add also a test for the tooltip with `await assert.tooltip...`. See doc [here](https://storybook.upfluence.co/?path=/docs/helpers-modifiers-modifiers-tooltip-testing--page)

tests/integration/components/o-s-s/attribute-text-test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    test('it renders a dash as value when no value is specified', async function (assert) {
+      await render(hbs`<OSS::Attribute::Text @label="Hello" />`);
+      assert.dom('.oss-attribute--text').exists();
+      assert.dom('.oss-attribute--text--value').hasText('-');
+    });
+
+    test('it copies the value', async function (assert) {
+      const writeTextStub = sinon.stub(navigator.clipboard, 'writeText').resolves();
```
JulienVannier66, line 34: Don't forget to call `restore()` on `writeTextStub` at the end of the test, otherwise you will break other tests (it is the case here)

tests/integration/components/o-s-s/attribute-text-test.ts
Miexil: This one is fine IMO, maybe change the label to clarify:
`It renders when the @value is undefined` (can be done with the empty state)

addon/components/o-s-s/attribute-text.ts
OwenCoogan: Here I need to check for undefined anyways because this.args.value is possibly undefined. but I reused the isBlank method for sure. 
Same for the copyable getter, had to juggle with the fact that it might be undefined it's a mix of what you suggested & avoiding typecheck errors

app/styles/atoms/attribute-text.less
OwenCoogan: Done but I'm guessing I'm going to have some merge issues when we get to your PR , might need to pass yours first & then I'll manage conflicts after your merge

addon/components/o-s-s/attribute-text.ts
OwenCoogan: Managed with the method used by @Miexil after taking a look at his PR

tests/integration/components/o-s-s/attribute-text-test.ts
OwenCoogan: Cut tests into a render module & a copy module & added required tests

addon/components/o-s-s/attribute-text.hbs
```diff
@@ -0,0 +1,20 @@
...
+     {{on "mouseleave" (fn (mut this.displayCopyBtn) false)}}
+     data-control-name="attribute-country">
+  <div class="oss-attribute--label">
+    <span>{{this.args.label}}</span>
+    {{#if this.args.tooltip}}
+      <OSS::Icon @icon="far fa-info-circle" class="oss-attribute--info" @inline={{true}}
+                 {{enable-tooltip title=this.args.tooltip placement="top"}} />
+    {{/if}}
+  </div>
+  <div class="oss-attribute--value">
```
phndiaye, line 12: `__value` (and `__label`) before per BEM syntax

addon/components/o-s-s/attribute-text.hbs
```diff
@@ -0,0 +1,20 @@
...
+    <span>{{this.value}}</span>
+  </div>
+  {{#if this.isCopyable}}
+    <div class="oss-attribute--copy">
```
phndiaye, line 16: and `__copy` here

addon/components/o-s-s/attribute-text.ts
```diff
@@ -0,0 +1,28 @@
...
+export default class OSSAttributeText extends Component<OSSAttributeTextArgs> {
+  @tracked displayCopyBtn = false;
+  @tracked copyable = this.args.copyable ?? true;
+  constructor(owner: unknown, args: OSSAttributeTextArgs) {
+    super(owner, args);
+    assert(`[component][OSS::Attribute::Text] label is required `, this.args.label);
+  }
+
+  get value(): string {
+    return isBlank(this.args.value) ? '-' : this.args.value ?? '-';
```
phndiaye, line 22: i don't think the `?? '—'` at the end is necessary as `isBlank(undefined)` returns true

app/styles/atoms/attribute.less
phndiaye, line 1: seems like you have a lot of common code w/ another PR from @Miexil 🤔 same comments above about BEM syntax should reflect here then :)

addon/components/o-s-s/attribute-text.stories.js
```diff
@@ -0,0 +1,73 @@
...
+      table: {
+        type: {
+          summary: 'string'
+        },
+        defaultValue: { summary: 'undefined' }
+      },
+      control: { type: 'text' }
+    },
+    value: {
+      type: { required: false },
```
JulienVannier66, line 19: The default value of the `required` is false, I think you can remove it when it `false`

addon/components/o-s-s/attribute-text.ts
```diff
@@ -0,0 +1,28 @@
...
+  label: string;
+  value?: string;
+  copyable?: boolean;
+  tooltip?: string;
+}
+
+export default class OSSAttributeText extends Component<OSSAttributeTextArgs> {
+  @tracked displayCopyBtn = false;
+  @tracked copyable = this.args.copyable ?? true;
+  constructor(owner: unknown, args: OSSAttributeTextArgs) {
```
JulienVannier66, line 16: Add space between tracked and constructor :smiley:

addon/components/o-s-s/attribute-text.ts
JulienVannier66:
```suggestion
    return isBlank(this.args.value) ? '-' : this.args.value;
```

app/styles/atoms/attribute.less
```diff
@@ -0,0 +1,36 @@
+.oss-attribute {
+  min-height: 24px;
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  justify-content: start;
+  padding-left: var(--spacing-px-9);
+  padding-right: var(--spacing-px-9);
+  gap: 9px;
```
JulienVannier66, line 9:
```suggestion
  gap: var(--spacing-px-9);
```

app/styles/atoms/attribute.less
```diff
@@ -0,0 +1,36 @@
...
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    justify-content: start;
+    gap: 6px;
+    &--info {
+      align-items: center;
+      cursor: pointer;
+    }
+  }
```
JulienVannier66, line 23:
```suggestion
  transition: all 0.25s ease-in-out;
  
  &--label {
    color: var(--color-gray-400);
    display: flex;
    flex-direction: row;
    align-items: center;
    justify-content: start;
    gap: var(--spacing-px-6);
    
    &--info {
      align-items: center;
      cursor: pointer;
    }
  }
```
Add space and use upfluence var :pray:

tests/integration/components/o-s-s/attribute-text-test.ts
```diff
@@ -0,0 +1,93 @@
...
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+import sinon from 'sinon';
+import click from '@ember/test-helpers/dom/click';
+import triggerEvent from '@ember/test-helpers/dom/trigger-event';
+
+module('Integration | Component | o-s-s/attribute-text', function (hooks) {
+  setupRenderingTest(hooks);
+  module('Default behavior', function () {
```
JulienVannier66, line 11: Add space between `setupRenderingTest` and `module`

tests/integration/components/o-s-s/attribute-text-test.ts
```diff
@@ -0,0 +1,93 @@
...
+        assert.dom('.oss-attribute--value').hasText('-');
+      });
+    });
+
+    module('Copy action', function (hooks) {
+      hooks.beforeEach(function () {
+        this.textForCopy = 'copied value';
+        this.displayCopyBtn = true;
+      });
+      test('the text is copyable by default', async function (assert) {
```
JulienVannier66, line 50: And here too :smiley:

tests/integration/components/o-s-s/attribute-text-test.ts
```diff
@@ -0,0 +1,93 @@
...
+        await assert.tooltip('.oss-attribute--info').hasTitle('Tooltip');
+      });
+
+      test('it renders a dash as value when no value is specified', async function (assert) {
+        await render(hbs`<OSS::Attribute::Text @label="Hello" />`);
+        assert.dom('.oss-attribute').exists();
+        assert.dom('.oss-attribute--value').hasText('-');
+      });
+
+      test('it renders a dash as value when value length is 0', async function (assert) {
```
JulienVannier66, line 38:
```suggestion
      test('it renders a dash as value when value is empty', async function (assert) {
```

tests/integration/components/o-s-s/attribute-text-test.ts
```diff
@@ -0,0 +1,93 @@
...
+        assert.dom('.oss-attribute--copy').exists();
+      });
+
+      test('The text is not copyable (no copy button) if the argument is falsy', async function (assert) {
+        await render(hbs`<OSS::Attribute::Text @label="Hello" @value={{this.textForCopy}} @copyable={{false}} />`);
+        await triggerEvent('.oss-attribute', 'mouseenter');
+        assert.dom('.oss-attribute--copy').doesNotExist();
+      });
+
+      test("The text is not copyable if text value isn't specified or the length is 0", async function (assert) {
```
JulienVannier66, line 69:
```suggestion
      test("The text is not copyable if text value isn't specified", async function (assert) {
```

tests/integration/components/o-s-s/attribute-text-test.ts
```diff
@@ -0,0 +1,93 @@
...
+        await render(hbs`<OSS::Attribute::Text @label="Hello" @copyable={{true}} />`);
+        await triggerEvent('.oss-attribute', 'mouseenter');
+        assert.dom('.oss-copy--inline').doesNotExist();
+      });
+
+      test('The text is not copyable if text value is specified but the length is 0', async function (assert) {
```
JulienVannier66, line 75:
```suggestion
      test('The text is not copyable if text value is empty', async function (assert) {
```

addon/components/o-s-s/attribute-text.ts
OwenCoogan: TS doesn't like it when I don't check for it @phndiaye  😢

app/styles/atoms/attribute.less
```diff
@@ -1,5 +1,6 @@
 .oss-attribute {
   height: 24px;
+  min-height: 24px;
```
Miexil, line 3: is this really needed ? 🤔

addon/components/o-s-s/attribute-text.hbs
```diff
@@ -0,0 +1,20 @@
...
+  <div class="oss-attribute__label">
+    <span>{{this.args.label}}</span>
+    {{#if this.args.tooltip}}
+      <OSS::Icon @icon="far fa-info-circle" class="oss-attribute__info" @inline={{true}}
+                 {{enable-tooltip title=this.args.tooltip placement="top"}} />
+    {{/if}}
+  </div>
+  <div class="oss-attribute__value">
+    <span>{{this.value}}</span>
+  </div>
```
Miexil, line 14: can directly be:
```
<span class="oss-attribute__value">{{this.value}}</span>
```

tests/integration/components/o-s-s/attribute-text-test.ts
```diff
@@ -0,0 +1,95 @@
...
+    module('Copy action', function (hooks) {
+      hooks.beforeEach(function () {
+        this.textForCopy = 'copied value';
+        this.displayCopyBtn = true;
+      });
+
+      test('the text is copyable by default', async function (assert) {
+        await render(hbs`<OSS::Attribute::Text @label="Hello" @value={{this.textForCopy}} />`);
+
+        await triggerEvent('.oss-attribute', 'mouseenter');
```
Miexil, line 55: I wanted to test that as well but the read in multiple locations from the ember community that `mouseenter` & `mouseleave` were unsafe events and thus didn't worked with `triggerEvent`. IT actually works ? 😱 
I'll be updating the Attribute::Country later then 👍

tests/integration/components/o-s-s/attribute-text-test.ts
```diff
@@ -0,0 +1,95 @@
...
+        assert.dom('.oss-copy--inline').doesNotExist();
+      });
+
+      test('The text is not copyable if text value is specified but the value is empty', async function (assert) {
+        await render(hbs`<OSS::Attribute::Text @label="Hello" @value="" @copyable={{true}} />`);
+        await triggerEvent('.oss-attribute', 'mouseenter');
+        assert.dom('.oss-copy--inline').doesNotExist();
+      });
+
+      test('The copy action is successfully called', async function (assert) {
```
Miexil, line 83: This is already tested at the `OSS::Copy` component level; no need to duplicate here :)

app/styles/atoms/attribute.less
OwenCoogan: Not sure depending on the use-case . I'd keep it simply because there is no impact if they are displayed in a column like in the designs but if we display them in a row , without this they start overflowing if the label or the value are too long. Just a small adjustment I added when checking the display on mobile.

Miexil: Then let's go for the trio ^^, can you add `max-height: 24px` ?

Miexil (approved):

JulienVannier66 (approved):

PR #321, opened 2023-10-26
Added: OSS::Attribute::Country component + tests

Miexil:
### What does this PR do?
Added: OSS::Attribute::Country component + tests
Also defines base classes for other Attribute components

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/attribute/country.hbs
+++ b/addon/components/o-s-s/attribute/country.hbs
@@ -2,13 +2,13 @@
      {{on "mouseenter" (fn (mut this.displayCopyBtn) true)}}
      {{on "mouseleave" (fn (mut this.displayCopyBtn) false)}}
      data-control-name="attribute-country" ...attributes>
-  <div class="oss-attribute--label">
+  <div class="oss-attribute__label">
     <span>{{t "oss-components.attribute.country"}}</span>
     {{#if (not-eq this.countryName "-")}}
       <div class="fflag fflag-{{@countryCode}} ff-round ff-sm"></div>
     {{/if}}
   </div>
-  <span class="oss-attribute--value">
+  <span class="oss-attribute__value">
     {{this.countryName}}
   </span>
   {{#if (and this.displayCopyBtn (not-eq this.countryName "-"))}}
--- a/addon/components/o-s-s/attribute/country.hbs
+++ b/addon/components/o-s-s/attribute/country.hbs
@@ -1,0 +1,17 @@
+<div class="oss-attribute"
+     {{on "mouseenter" (fn (mut this.displayCopyBtn) true)}}
+     {{on "mouseleave" (fn (mut this.displayCopyBtn) false)}}
+     data-control-name="attribute-country" ...attributes>
+  <div class="oss-attribute--label">
+    <span>{{t "oss-components.attribute.country"}}</span>
+    {{#if (not-eq this.countryName "-")}}
+      <div class="fflag fflag-{{@countryCode}} ff-round ff-sm"></div>
+    {{/if}}
+  </div>
+  <span class="oss-attribute--value">
+    {{this.countryName}}
+  </span>
+  {{#if (and this.displayCopyBtn (not-eq this.countryName "-"))}}
+    <OSS::Copy @value={{this.countryName}} @inline="true" />
+  {{/if}}
+</div>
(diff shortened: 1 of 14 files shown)


addon/components/o-s-s/attribute/country.hbs
```diff
@@ -0,0 +1,17 @@
+<div class="oss-attribute"
+     {{on "mouseenter" (fn (mut this.displayCopyBtn) true)}}
+     {{on "mouseleave" (fn (mut this.displayCopyBtn) false)}}
+     data-control-name="attribute-country" ...attributes>
```
phndiaye, line 4: should we really put data-control-name here instead of letting the "context" set one ? maybe in the page it will be used it will more like "influencer-country" or "shipping-address-country" ?

addon/components/o-s-s/copy.hbs
```diff
@@ -1,7 +1,9 @@
 {{#if this.inline}}
   <OSS::Icon @icon="fa-copy" class="oss-copy--inline" {{on "click" this.copy}}
-             {{enable-tooltip placement="top" title=(t "oss-components.copy.tooltip") trigger="hover"}} />
+             {{enable-tooltip placement="top" title=(t "oss-components.copy.tooltip") trigger="hover"}}
+             data-control-name="copy-content-button" />
```
phndiaye, line 4: same comment as above about data-control-name, to me it's the role of the context to display it. we should add a `...attributes` here however 🤔  missed it during the review :(

app/styles/atoms/attribute.less
```diff
@@ -0,0 +1,29 @@
...
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  justify-content: start;
+  padding-left: var(--spacing-px-9);
+  padding-right: var(--spacing-px-9);
+  gap: var(--spacing-px-9);
+  border-radius: var(--border-radius-sm);
+
+  &--label {
```
phndiaye, line 12: `&__label` for bem :p

app/styles/atoms/attribute.less
```diff
@@ -0,0 +1,29 @@
...
+    display: flex;
+    color: var(--color-gray-400);
+    flex-direction: row;
+    align-items: center;
+    justify-content: start;
+    gap: var(--spacing-px-6);
+  }
+
+  &--value {
```
phndiaye, line 21: `&__value` for bem :p

addon/components/o-s-s/attribute/country.hbs
Miexil: I was thinking better to have a default one at overwrite it if need be, no ? (unless we can't overwrite with the splattributes ? ^^)

addon/components/o-s-s/copy.hbs
Miexil: I'll add the splattributes here as well.
for the data-control-name, I say we're better with a default value that can be overwritten. But if someone else prefers to enforce them via the parent, I'll remove them !

app/styles/atoms/attribute.less
```diff
@@ -0,0 +1,29 @@
...
+
+  &__value {
+    color: var(--color-gray-900);
+    flex: 1;
+  }
+}
+
+.oss-attribute:hover {
+  background: var(--color-gray-50);
+}
```
JulienVannier66, line 29: Nit: You can also add this in the `.oss-attribute` directly
```suggestion
&:hover {
  background: var(--color-gray-50);
}
```

translations/en-us.yaml
```diff
@@ -1,6 +1,6 @@
 oss-components:
   upf-table:
-    pagination: "{from} to {to} out of {total} {item_name}"
+    pagination: '{from} to {to} out of {total} {item_name}'
```
JulienVannier66, line 3: Why changing quote type?

addon/components/o-s-s/copy.hbs
JulienVannier66: I don't think it is a bad idea to set it here, I'm with Max on this one

phndiaye (approved):

translations/en-us.yaml
Miexil: auto linting from something on my machine ^^
Do you want me to rollback the changes ?

JulienVannier66: If it is work, it is ok for me

JulienVannier66 (approved):
LGTM

PR #322, opened 2023-10-26
New OSS::AttributesPanel component

JulienVannier66:
### What does this PR do?

Add new OSS::AttributesPanel component
Add tests & documentation

Related to: https://linear.app/upfluence/issue/ENG-1555/[oss-components]-new-ossattributespanel-component

Peek 2023-10-26 17-40

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/attributes-panel.hbs
+++ b/addon/components/o-s-s/attributes-panel.hbs
@@ -29,9 +29,10 @@
         <div class="fx-col fx-gap-px-18">
           {{yield to="edition-mode"}}
           <div class="fx-row fx-gap-px-18 fx-malign-end">
-            <OSS::Button @label="Cancel" {{on "click" this.onCancel}} data-control-name="attributes-panel-cancel-button" />
-            <OSS::Button @skin="primary" @label="Save" @loading={{this.isLoading}} {{on "click" this.onSave}}
-                         data-control-name="attributes-panel-save-button" />
+            <OSS::Button @label={{t "oss-components.attributes_panel.cancel"}} {{on "click" this.onCancel}}
+                         data-control-name="attributes-panel-cancel-button" />
+            <OSS::Button @skin="primary" @label={{t "oss-components.attributes_panel.save"}} @loading={{this.isLoading}}
+                         {{on "click" this.onSave}} data-control-name="attributes-panel-save-button" />
           </div>
         </div>
       {{/if}}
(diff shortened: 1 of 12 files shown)


tests/integration/components/o-s-s/attributes-panel-test.ts
```diff
@@ -0,0 +1,175 @@
...
+  module('for view mode', (hooks) => {
+    hooks.beforeEach(async function () {
+      await render(hbs`
+        <OSS::AttributesPanel @icon={{this.icon}} @title={{this.title}} @onSave={{this.onSave}} @onEdit={{this.onEdit}}>
+          <:view-mode>
+            <div class="custom-view-mode">View mode</div>
+          </:view-mode>
+        </OSS::AttributesPanel>
+      `);
+    });
```
Miexil, line 53: I think @OwenCoogan started a trend with component rendering in the hooks.

as discussed [in this slack thread](https://upfluence.slack.com/archives/C02FBD3QAQ0/p1698221992007299) ; I think we should keep the methods in the hooks for service / variable initialization and not have renders here.

If we want to spare some lines of code, I think we should better create a private function with a proper naming which makes the render. [Here's an example of that](https://github.com/upfluence/ember-influencer/blob/master/tests/integration/components/discount-code/edition-test.ts#L159)

IMO, we are losing readability in the tests when moving the render inside the hooks ; it might not be a big deal here because we only have 2 test cases but imagine if we had 10+ test cases without any notion of which component is rendered inside each test ?

phndiaye: +1.

If I can add something, I would say that given that most of the time we try to have a Given/When/Then way of writing the tests, this means that we are doing Given and When together :/ So yeaa, keeping hooks to "setting up the context" seems better to me. and the extra private method is not much to write AND can even take extra args for even more setup if needed.

addon/components/o-s-s/attributes-panel.hbs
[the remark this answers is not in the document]
Miexil: ( clear thoughts after a good night's sleep ^^ )
Actually if you want to keep this outside it also makes sense, plus the name has AttributesPanel (with an `s`) and holds all of the attributes (& they will be located in o-s-s/attribute/x)

So you can omit my comment 😇

tests/integration/components/o-s-s/attributes-panel-test.ts
JulienVannier66: Ok ok no problem, I will update that

Miexil (approved):

OwenCoogan (approved):

PR #331, opened 2023-11-09
Added OSS::Attribute::Rating

OwenCoogan:
### What does this PR do?
Added OSS::Attribute::Rating with tests & storybook story

Related to: #
https://linear.app/upfluence/issue/ENG-1561/[oss-components]-new-ossattributerating-component

https://www.loom.com/share/ec0a0f3dfbc244308b5ddbe879c6a2bd?sid=ddb551be-65ee-4e25-a219-51d515d68065

Capture d’écran 2023-11-08 à 17 30 22

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/attribute/rating.hbs
+++ b/addon/components/o-s-s/attribute/rating.hbs
@@ -4,15 +4,10 @@
   </:label>
   <:value>
     {{#if @rating}}
-      <OSS::StarRating @rating={{@rating}}
-                       @totalStars={{5}}
-                       @activeColor="yellow"
-                       @passiveColor="yellow"
-                       @passiveStyle="regular"
-                       @onChange={{this.onClick}} />
+      <OSS::StarRating @rating={{@rating}} @totalStars={{5}} @activeColor="yellow"
+                       @passiveColor="yellow" @passiveStyle="regular" @onChange={{this.onClick}} />
     {{else}}
       -
     {{/if}}
-
   </:value>
 </OSS::Attribute::Base>
--- a/addon/components/o-s-s/attribute/rating.hbs
+++ b/addon/components/o-s-s/attribute/rating.hbs
@@ -9,7 +9,7 @@
                        @activeColor="yellow"
                        @passiveColor="yellow"
                        @passiveStyle="regular"
-                       @onChange={{@onChange}} />
+                       @onChange={{this.onClick}} />
     {{else}}
       -
     {{/if}}
--- a/addon/components/o-s-s/attribute/rating.hbs
+++ b/addon/components/o-s-s/attribute/rating.hbs
@@ -1,0 +1,18 @@
+<OSS::Attribute::Base data-control-name="attribute-rating" ...attributes>
+  <:label>
+    <span>{{@label}}</span>
+  </:label>
+  <:value>
+    {{#if @rating}}
+      <OSS::StarRating @rating={{@rating}}
+                       @totalStars={{5}}
+                       @activeColor="yellow"
+                       @passiveColor="yellow"
+                       @passiveStyle="regular"
+                       @onChange={{@onChange}} />
+    {{else}}
+      -
+    {{/if}}
+
+  </:value>
+</OSS::Attribute::Base>
--- a/addon/components/o-s-s/attribute/rating.hbs
+++ b/addon/components/o-s-s/attribute/rating.hbs
@@ -1,11 +1,11 @@
-<OSS::Attribute::Base data-control-name="attribute-rating" ...attributes>
+<OSS::Attribute::Base class="oss-attribute--no-hover-effect" data-control-name="attribute-rating" ...attributes>
   <:label>
     <span>{{@label}}</span>
   </:label>
   <:value>
     {{#if @rating}}
       <OSS::StarRating @rating={{@rating}} @totalStars={{5}} @activeColor="yellow"
-                       @passiveColor="yellow" @passiveStyle="regular" @onChange={{this.onClick}} />
+                       @passiveColor="yellow" @passiveStyle="regular" />
     {{else}}
       -
     {{/if}}
--- a/addon/components/o-s-s/attribute/rating.hbs
+++ b/addon/components/o-s-s/attribute/rating.hbs
@@ -5,7 +5,7 @@
   <:value>
     {{#if @rating}}
       <OSS::StarRating @rating={{@rating}} @totalStars={{5}} @activeColor="yellow"
-                       @passiveColor="yellow" @passiveStyle="regular" />
+                       @passiveColor="yellow" @passiveStyle="regular" @disablePointerCursor={{true}} />
     {{else}}
       -
     {{/if}}
--- a/addon/components/o-s-s/attribute/rating.stories.js
+++ b/addon/components/o-s-s/attribute/rating.stories.js
@@ -26,7 +26,6 @@ export default {
       control: { type: 'number' }
     },
     onChange: {
-      type: { required: false },
       description: 'The action triggered when a rating star is clicked',
       table: {
         category: 'Actions',
@@ -54,8 +53,7 @@ const defaultArgs = {
 const BasicUsageTemplate = (args) => ({
   template: hbs`
     <div style="padding: 12px; background: white">
-       <OSS::Attribute::Rating
-                                @label="Rating"
+       <OSS::Attribute::Rating  @label={{this.label}}
                                 @rating={{this.rating}}
                                 @onChange={{this.onChange}} />
     </div>
--- a/addon/components/o-s-s/attribute/rating.stories.js
+++ b/addon/components/o-s-s/attribute/rating.stories.js
@@ -24,13 +24,6 @@ export default {
         defaultValue: { summary: 'undefined' }
       },
       control: { type: 'number' }
-    },
-    onChange: {
-      description: 'The action triggered when a rating star is clicked',
-      table: {
-        category: 'Actions',
-        type: { summary: 'onChange(index: number): void' }
-      }
     }
   },
   parameters: {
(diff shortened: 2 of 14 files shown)


addon/components/o-s-s/attribute/rating.hbs
```diff
@@ -0,0 +1,18 @@
...
+    <span>{{@label}}</span>
+  </:label>
+  <:value>
+    {{#if @rating}}
+      <OSS::StarRating @rating={{@rating}}
+                       @totalStars={{5}}
+                       @activeColor="yellow"
+                       @passiveColor="yellow"
+                       @passiveStyle="regular"
+                       @onChange={{this.onClick}} />
```
Miexil, line 12: maybe spread out the variables on line 1 & 2 ^^

addon/components/o-s-s/attribute/rating.ts
```diff
@@ -0,0 +1,23 @@
...
+export default class OSSAttributeRating extends Component<OSSAttributeRatingArgs> {
+  constructor(owner: unknown, args: OSSAttributeRatingArgs) {
+    super(owner, args);
+    assert('[component][OSS::Attribute::Rating] @label is required', typeof args.label === 'string');
+  }
+  @action
+  onClick(rating: number) {
+    if (this.args.onChange) {
+      this.args.onChange(rating);
+    }
```
JulienVannier66, line 21: Nit: `this.args.onChange?.(rating);`

addon/components/o-s-s/star-rating.ts
```diff
@@ -1,11 +1,14 @@
 import Component from '@glimmer/component';
 import { assert } from '@ember/debug';
+import { action } from '@ember/object';
 
-interface OSSStarRatingArgs {
+export interface OSSStarRatingArgs {
```
JulienVannier66, line 5: Why export it?

OwenCoogan: For consistency between onChange event in the OSSStarRatingArgs interface &  the one in starRating. 
The interface in Attribute::Rating now has `onChange?: OSSStarRatingArgs['onChange'];` .

OwenCoogan: Since we don't use onChange on Attr::Rating, I removed the export @JulienVannier66

addon/components/o-s-s/attribute/rating.stories.js
```diff
@@ -0,0 +1,65 @@
...
+      },
+      control: { type: 'number' }
+    },
+    onChange: {
+      description: 'The action triggered when a rating star is clicked',
+      table: {
+        category: 'Actions',
+        type: { summary: 'onChange(index: number): void' }
+      }
+    }
```
JulienVannier66, line 34: Remove it from the doc, not used anymore

addon/components/o-s-s/attribute/rating.stories.js
```diff
@@ -0,0 +1,65 @@
...
+      },
+      iframeHeight: 150
+    }
+  }
+};
+
+const defaultArgs = {
+  label: 'Label',
+  rating: 3,
+  onChange: (index) => alert('clicked on ' + index)
```
JulienVannier66, line 50: Same here and in the `BasicUsageTemplate`

tests/dummy/app/controllers/application.js
```diff
@@ -218,6 +219,10 @@ export default class ApplicationController extends Controller {
     this.mockUploader = owner.lookup('service:mock-uploader');
   }
 
+  @action onRatingClick(rating) {
```
JulienVannier66, line 222:
```suggestion
  @action
  onRatingClick(rating) {
```
Set `onRatingClick` on a new line

JulienVannier66 (approved):
Otherwise LGTM

addon/components/o-s-s/star-rating.ts
```diff
@@ -38,18 +45,37 @@ export default class OSSStarRating extends Component<OSSStarRatingArgs> {
   }
 
   get activeColorClass(): string {
-    return `color-${this.args.activeColor || 'yellow'}`;
+    return `${this.args.onChange && 'oss-star-rating'} color-${this.args.activeColor || 'yellow'}`;
   }
 
   get passiveColorClass(): string {
-    return `color-${this.args.passiveColor || 'grey'}`;
+    return `${this.args.onChange && 'oss-star-rating'} color-${this.args.passiveColor || 'grey'}`;
```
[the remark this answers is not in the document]
Miexil: oh, I saw the less code, so this is to prevent having a pointer cursor when using the Attribute::Rating, right ?
I think it would be better to restore the orginal getters here and to add a `splattribute` to the `o-s-s/star-rating.hbs` at the top level.
This way, from the `OSS::Attribute::Rating` you can pass a new class, something like "no-hover-pointer" and deactivate it from the parent.

tests/integration/components/o-s-s/star-rating-test.ts
```diff
@@ -22,7 +22,23 @@ module('Integration | Component | o-s-s/star-rating', function (hooks) {
                                         @passiveColor={{this.passiveColor}} />`);
       assert.dom('.fas').hasClass(`color-${starColor}`);
     });
+
+    test('Passing @passiveStyle and @passiveColor applies the correct classes to the component', async function (assert) {
```
Miexil, line 26: passiveColor is already tested in the loop above.
Either add your `passiveStyle` test to the existing test or move this test out of the `for` loop and test for both `regular` and `solid` styles

tests/integration/components/o-s-s/star-rating-test.ts
```diff
@@ -22,7 +22,23 @@ module('Integration | Component | o-s-s/star-rating', function (hooks) {
...
+      this.activeColor = starColor;
+      this.passiveColor = starColor;
+      await render(hbs`<OSS::StarRating @rating={{5}} @totalStars={{8}} @passiveStyle='regular'
+                                        @activeColor={{this.activeColor}}
+                                        @passiveColor={{this.passiveColor}} />`);
+      assert.dom('.far').hasClass(`color-${starColor}`);
+    });
   }
+  test('the number of stars displayed is consistent with the provided @rating', async function (assert) {
```
Miexil, line 35: missing line break. And capital `The number of`

addon/components/o-s-s/star-rating.ts
```diff
@@ -1,11 +1,15 @@
...
+import { action, set } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
 
 interface OSSStarRatingArgs {
   rating: number;
   totalStars: number;
   activeColor: StarColor;
   passiveColor: StarColor;
+  passiveStyle?: 'solid' | 'regular';
+  onChange?: (rating: OSSStarRatingArgs['rating']) => void;
```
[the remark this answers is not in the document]
OwenCoogan: I think you're right @Miexil . This make a lot more sense when doing it through an exported interface from some far away file, not that much when it's 4 lines from above & in the case of a simple number type, lisibility is better achieved through a number type. I did it when I was using the shared onchange event between Attr:Rating & OSS::StarRating, which made a bit more sense than in this case /

Miexil (approved):

addon/components/o-s-s/star-rating.ts
```diff
@@ -1,11 +1,18 @@
...
+type StarType = 'solid' | 'regular';
+type Star = { type: StarType };
 
 interface OSSStarRatingArgs {
   rating: number;
   totalStars: number;
   activeColor: StarColor;
   passiveColor: StarColor;
+  passiveStyle?: StarType;
+  onChange?: (rating: number) => void;
```
JulienVannier66, line 15:
```suggestion
  onChange?(rating: number): void;
```
This syntax is used instead

addon/components/o-s-s/star-rating.ts
```diff
@@ -45,11 +55,30 @@ export default class OSSStarRating extends Component<OSSStarRatingArgs> {
...
+  }
+
+  @action
+  onMouseLeave(): void {
+    this.stars = this.generateStarsArray();
   }
 
-  get passiveStars(): any[] {
-    return new Array(Math.max(0, this.args.totalStars - this.args.rating));
+  private generateStarsArray(): any[] {
```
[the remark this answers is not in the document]
Miexil: 2499c520934779 562f38e953dfc

JulienVannier66 (approved):
Otherwise LGTM

addon/components/o-s-s/star-rating.hbs
```diff
@@ -1,8 +1,8 @@
 <div class="star-rating fx-row">
-  {{#each this.activeStars}}
-    <OSS::Icon @style="solid" @icon="fa-star" class={{this.activeColorClass}} />
-  {{/each}}
-  {{#each this.passiveStars}}
-    <OSS::Icon @style="solid" @icon="fa-star" class={{this.passiveColorClass}} />
+  {{#each this.stars as |star index|}}
+    <OSS::Icon class="{{unless this.disablePointerCursor 'no-pointer-cursor'}} {{this.activeColorClass}}"
```
phndiaye: given there is an `activeColorClass` getter, why not not move the `disablePointerCursor`  handling there?
alternatively, you can leave the `cursor` change to a `role="button"` attribute on the element, meaning you can have something like (and no need for custom CSS either for cursor: auto or pointer anymore) :
```
<OSS::Icon ... role={{unless this.disablePointerCursor "button"}} />
```

Miexil: the usage of role sounds good, the first part does not seem compatible though (I think). In any case we're getting active and passive stars, it's just that when `disablePointerCursor` is enabled we don't want the stars to be clickable :)

phndiaye (approved):

addon/components/o-s-s/star-rating.hbs
OwenCoogan: @phndiaye that was what I did in the first place but we wanted to manage it from the parent & not make it dependent on the onChange argument to set it in the getter

PR #338, opened 2023-11-22
Added OSS::ProgressBar

OwenCoogan:
### What does this PR do?

It adds a new OSS::ProgressBar component

Related to: #
https://linear.app/upfluence/issue/ENG-509/get-rid-of-upf-progress-ossprogressbar
### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/progress-bar.hbs
+++ b/addon/components/o-s-s/progress-bar.hbs
@@ -5,7 +5,7 @@
     </span>
   {{/if}}
   <div class="oss-progress-bar__outer">
-    <div class={{concat "oss-progress-bar__inner"}}
+    <div class="oss-progress-bar__inner"
          role="progressbar" aria-valuenow={{@value}}
          aria-valuemin="0" aria-valuemax="100" style={{progressBarWidthStyle}}>
     </div>
--- a/addon/components/o-s-s/progress-bar.hbs
+++ b/addon/components/o-s-s/progress-bar.hbs
@@ -1,18 +1,18 @@
 <div class='oss-progress-bar'>
   {{#if @label}}
-    <span class='oss-progress-bar--label'>
+    <span class='oss-progress-bar__label'>
       {{@label}}
     </span>
   {{/if}}
-  <div class='oss-progress-bar--outter'>
+  <div class='oss-progress-bar__outer'>
     <div
-         class={{concat 'oss-progress-bar--inner ' this.computedStyles (concat ' oss-progress-bar--inner--' this.size)}}
+         class={{concat 'oss-progress-bar__inner ' this.computedStyles}}
          role='progressbar' aria-valuenow={{@value}}
-         aria-valuemin='0' aria-valuemax='100' style={{progressBarInnerStyle}}>
+         aria-valuemin='0' aria-valuemax='100' style={{progressBarWidthStyle}}>
     </div>
   </div>
   {{#if @displayValue}}
-    <span class='oss-progress-bar--value'>
+    <span class='oss-progress-bar__value'>
       {{@value}}%
     </span>
   {{/if}}
--- a/addon/components/o-s-s/progress-bar.hbs
+++ b/addon/components/o-s-s/progress-bar.hbs
@@ -1,14 +1,14 @@
-<div class='oss-progress-bar--wrapper'>
+<div class='oss-progress-bar'>
   {{#if @label}}
     <span class='oss-progress-bar--label'>
       {{@label}}
     </span>
   {{/if}}
-  <div class='oss-progress-bar'>
-    <div 
-      class={{concat 'oss-progress-bar--inner ' this.computedStyles (if @small ' oss-progress-bar--inner--small') }}
-      role='progressbar' aria-valuenow={{@value}}
-      aria-valuemin='0' aria-valuemax='100' style={{progressBarInnerStyle}}>
+  <div class='oss-progress-bar--outter'>
+    <div
+         class={{concat 'oss-progress-bar--inner ' this.computedStyles (concat ' oss-progress-bar--inner--' this.size)}}
+         role='progressbar' aria-valuenow={{@value}}
+         aria-valuemin='0' aria-valuemax='100' style={{progressBarInnerStyle}}>
     </div>
   </div>
   {{#if @displayValue}}
--- a/addon/components/o-s-s/progress-bar.hbs
+++ b/addon/components/o-s-s/progress-bar.hbs
@@ -1,18 +1,17 @@
-<div class={{concat 'oss-progress-bar ' this.computedSize}} ...attributes>
+<div class={{concat "oss-progress-bar " this.computedSize}} ...attributes>
   {{#if @label}}
-    <span class='oss-progress-bar__label'>
+    <span class="oss-progress-bar__label">
       {{@label}}
     </span>
   {{/if}}
-  <div class='oss-progress-bar__outer'>
-    <div
-         class={{concat 'oss-progress-bar__inner ' this.computedStyles}}
-         role='progressbar' aria-valuenow={{@value}}
-         aria-valuemin='0' aria-valuemax='100' style={{progressBarWidthStyle}}>
+  <div class="oss-progress-bar__outer">
+    <div class={{concat "oss-progress-bar__inner " this.computedStyles}}
+         role="progressbar" aria-valuenow={{@value}}
+         aria-valuemin="0" aria-valuemax="100" style={{progressBarWidthStyle}}>
     </div>
   </div>
   {{#if @displayValue}}
-    <span class='oss-progress-bar__value'>
+    <span class="oss-progress-bar__value">
       {{@value}}%
     </span>
   {{/if}}
--- a/addon/components/o-s-s/progress-bar.hbs
+++ b/addon/components/o-s-s/progress-bar.hbs
@@ -1,8 +1,8 @@
 <div class="oss-progress-bar--wrapper">
   {{#if @label}}
-    <p class="oss-progress-bar--label">
+    <span class="oss-progress-bar--label">
       {{@label}}
-    </p>
+    </span>
   {{/if}}
   <div class="oss-progress-bar">
     <div 
@@ -12,8 +12,8 @@
     </div>
   </div>
   {{#if @valueIsVisible}}
-    <p class="oss-progress-bar--value">
+    <span class="oss-progress-bar--value">
       {{@value}}%
-    </p>
+    </span>
   {{/if}}
 </div>
--- a/addon/components/o-s-s/progress-bar.hbs
+++ b/addon/components/o-s-s/progress-bar.hbs
@@ -1,18 +1,18 @@
-<div class="oss-progress-bar--wrapper">
+<div class='oss-progress-bar--wrapper'>
   {{#if @label}}
-    <span class="oss-progress-bar--label">
+    <span class='oss-progress-bar--label'>
       {{@label}}
     </span>
   {{/if}}
-  <div class="oss-progress-bar">
+  <div class='oss-progress-bar'>
     <div 
-      class={{concat "oss-progress-bar--inner " this.computedStyles (if @small " oss-progress-bar--inner--small") }}
-      role="progressbar" aria-valuenow={{@value}}
+      class={{concat 'oss-progress-bar--inner ' this.computedStyles (if @small ' oss-progress-bar--inner--small') }}
+      role='progressbar' aria-valuenow={{@value}}
       aria-valuemin='0' aria-valuemax='100' style={{progressBarInnerStyle}}>
     </div>
   </div>
-  {{#if @valueIsVisible}}
-    <span class="oss-progress-bar--value">
+  {{#if @displayValue}}
+    <span class='oss-progress-bar--value'>
       {{@value}}%
     </span>
   {{/if}}
--- a/addon/components/o-s-s/progress-bar.hbs
+++ b/addon/components/o-s-s/progress-bar.hbs
@@ -1,15 +1,16 @@
 <div class="oss-progress-bar--wrapper">
-  {{#if this.args.label}}
+  {{#if @label}}
     <p class="oss-progress-bar--label">
       {{@label}}
     </p>
   {{/if}}
   <div class="oss-progress-bar">
-    <div class={{concat "oss-progress-bar--inner " this.computedStyles }} aria-valuenow={{@value}}
+    <div class={{concat "oss-progress-bar--inner " this.computedStyles (if @small " oss-progress-bar--inner--small") }}
+         role="progressbar" aria-valuenow={{@value}}
          aria-valuemin='0' aria-valuemax='100' style={{progressBarInnerStyle}}>
     </div>
   </div>
-  {{#if this.args.valueIsVisible}}
+  {{#if @valueIsVisible}}
     <p class="oss-progress-bar--value">
       {{@value}}%
     </p>
--- a/addon/components/o-s-s/progress-bar.hbs
+++ b/addon/components/o-s-s/progress-bar.hbs
@@ -1,11 +1,11 @@
-<div class={{concat "oss-progress-bar " this.computedSize}} ...attributes>
+<div class={{this.computedStyles}} ...attributes>
   {{#if @label}}
     <span class="oss-progress-bar__label">
       {{@label}}
     </span>
   {{/if}}
   <div class="oss-progress-bar__outer">
-    <div class={{concat "oss-progress-bar__inner " this.computedStyles}}
+    <div class={{concat "oss-progress-bar__inner"}}
          role="progressbar" aria-valuenow={{@value}}
          aria-valuemin="0" aria-valuemax="100" style={{progressBarWidthStyle}}>
     </div>
--- a/addon/components/o-s-s/progress-bar.hbs
+++ b/addon/components/o-s-s/progress-bar.hbs
@@ -1,4 +1,4 @@
-<div class='oss-progress-bar'>
+<div class={{concat 'oss-progress-bar ' this.computedSize}} ...attributes>
   {{#if @label}}
     <span class='oss-progress-bar__label'>
       {{@label}}
--- a/addon/components/o-s-s/progress-bar.hbs
+++ b/addon/components/o-s-s/progress-bar.hbs
@@ -5,9 +5,10 @@
     </p>
   {{/if}}
   <div class="oss-progress-bar">
-    <div class={{concat "oss-progress-bar--inner " this.computedStyles (if @small " oss-progress-bar--inner--small") }}
-         role="progressbar" aria-valuenow={{@value}}
-         aria-valuemin='0' aria-valuemax='100' style={{progressBarInnerStyle}}>
+    <div 
+      class={{concat "oss-progress-bar--inner " this.computedStyles (if @small " oss-progress-bar--inner--small") }}
+      role="progressbar" aria-valuenow={{@value}}
+      aria-valuemin='0' aria-valuemax='100' style={{progressBarInnerStyle}}>
     </div>
   </div>
   {{#if @valueIsVisible}}
--- a/addon/components/o-s-s/progress-bar.hbs
+++ b/addon/components/o-s-s/progress-bar.hbs
@@ -1,0 +1,17 @@
+<div class="oss-progress-bar--wrapper">
+  {{#if this.args.label}}
+    <p class="oss-progress-bar--label">
+      {{@label}}
+    </p>
+  {{/if}}
+  <div class="oss-progress-bar">
+    <div class={{concat "oss-progress-bar--inner " this.computedStyles }} aria-valuenow={{@value}}
+         aria-valuemin='0' aria-valuemax='100' style={{progressBarInnerStyle}}>
+    </div>
+  </div>
+  {{#if this.args.valueIsVisible}}
+    <p class="oss-progress-bar--value">
+      {{@value}}%
+    </p>
+  {{/if}}
+</div>
--- a/addon/components/o-s-s/progress-bar.stories.js
+++ b/addon/components/o-s-s/progress-bar.stories.js
@@ -44,7 +44,7 @@ export default {
         type: {
           summary: ProgressBarSizes.join('|')
         },
-        defaultValue: { summary: 'sm' }
+        defaultValue: { summary: 'lg' }
       },
       options: ProgressBarSizes,
       control: { type: 'select' }
@@ -75,7 +75,7 @@ const defaultArgs = {
   label: 'Hello',
   displayValue: true,
   skin: 'primary',
-  size: 'sm'
+  size: 'lg'
 };
 
 const BasicUsageTemplate = (args) => ({
--- a/addon/components/o-s-s/progress-bar.stories.js
+++ b/addon/components/o-s-s/progress-bar.stories.js
@@ -27,7 +27,7 @@ export default {
       },
       control: { type: 'text' }
     },
-    valueIsVisible: {
+    displayValue: {
       description: 'Boolean to check if @value should be displayed on the right side of the bar',
       table: {
         type: {
@@ -38,7 +38,7 @@ export default {
       control: { type: 'boolean' }
     },
     small: {
-      description: 'Boolean to adjust height of progress bar',
+      description: 'If true, the height will be 4px. The default height is 10px',
       table: {
         type: {
           summary: 'boolean'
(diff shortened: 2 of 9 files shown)


OwenCoogan:
https://www.loom.com/share/0443b15a0f5342578a50a47ba29dc7a3?sid=ed3ac921-e0b1-4c78-8a93-ec9524b6f064

addon/components/o-s-s/progress-bar.hbs
```diff
@@ -0,0 +1,19 @@
+<div class="oss-progress-bar--wrapper">
+  {{#if @label}}
+    <p class="oss-progress-bar--label">
+      {{@label}}
+    </p>
```
JulienVannier66, line 5: Maybe use `span` instead of `p` here

addon/components/o-s-s/progress-bar.hbs
```diff
@@ -0,0 +1,19 @@
...
+    <div 
+      class={{concat "oss-progress-bar--inner " this.computedStyles (if @small " oss-progress-bar--inner--small") }}
+      role="progressbar" aria-valuenow={{@value}}
+      aria-valuemin='0' aria-valuemax='100' style={{progressBarInnerStyle}}>
+    </div>
+  </div>
+  {{#if @valueIsVisible}}
+    <p class="oss-progress-bar--value">
+      {{@value}}%
+    </p>
```
JulienVannier66, line 17: Same here

addon/components/o-s-s/progress-bar.stories.js
```diff
@@ -0,0 +1,91 @@
...
+      },
+      control: { type: 'number' }
+    },
+    label: {
+      description: 'Displayed label',
+      table: {
+        type: {
+          summary: 'string'
+        },
+        defaultValue: { summary: 'Hello' }
```
JulienVannier66, line 26:
```suggestion
        defaultValue: { summary: 'undefined' }
```
The default value is `undefined` here

addon/components/o-s-s/progress-bar.stories.js
```diff
@@ -0,0 +1,91 @@
...
+      },
+      control: { type: 'text' }
+    },
+    valueIsVisible: {
+      description: 'Boolean to check if @value should be displayed on the right side of the bar',
+      table: {
+        type: {
+          summary: 'boolean'
+        },
+        defaultValue: { summary: true }
```
JulienVannier66, line 36:
```suggestion
        defaultValue: { summary: 'undefined }
```
Same here, no default value set by the component

addon/components/o-s-s/progress-bar.stories.js
```diff
@@ -0,0 +1,91 @@
...
+      },
+      control: { type: 'boolean' }
+    },
+    small: {
+      description: 'Boolean to adjust height of progress bar',
+      table: {
+        type: {
+          summary: 'boolean'
+        },
+        defaultValue: { summary: true }
```
JulienVannier66, line 46:
```suggestion
        defaultValue: { summary: 'undefined' }
```

addon/components/o-s-s/progress-bar.ts
```diff
@@ -0,0 +1,50 @@
+import Component from '@glimmer/component';
+import { assert } from '@ember/debug';
+
+enum ProgressBarSkins {
+  'primary',
+  'warning',
+  'success'
+}
```
JulienVannier66, line 8: With an `enum` here and not type definition like `type ProgressBarSkins = 'primary' | 'warning' | 'success'`

app/styles/atoms/progress-bar.less
```diff
@@ -0,0 +1,41 @@
+.oss-progress-bar--wrapper {
+  min-width: 220px;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  p {
+    font-weight: 200;
+    margin: 0 6px;
```
JulienVannier66, line 8:
```suggestion
    margin: 0 var(--spacing-px-6);
```

app/styles/atoms/progress-bar.less
```diff
@@ -0,0 +1,41 @@
...
+  }
+}
+
+.oss-progress-bar {
+  width: 100%;
+  border-radius: var(--border-radius-sm);
+  background-color: darken(@upf-gray-light, 8%);
```
JulienVannier66, line 15: We deprecated the use of darken, if the color doesn't exist in the design system, sync with William or set the hex value directly

tests/integration/components/o-s-s/progress-bar-test.ts
```diff
@@ -0,0 +1,95 @@
...
+    });
+  });
+
+  module('@value behaviour', function () {
+    test('It renders the progress bar inner with the correct width when @value is specified', async function (assert) {
+      await render(hbs`<OSS::ProgressBar @value={{this.checkedValue}} />`);
+
+      assert.dom('.oss-progress-bar').exists();
+      const innerBar = this.element.querySelector('.oss-progress-bar--inner');
+      if (innerBar) {
```
JulienVannier66, line 42: Why using a `if` here? Same below

tests/integration/components/o-s-s/progress-bar-test.ts
JulienVannier66, line 1: Maybe for a better visibility, you can group each test for each attribute in the same module. For example
```
module('for @small', function () {
  test('if the value is true, ...)
  test('if the value is false, ...)
}
```

JulienVannier66, line 1: Maybe for a better visibility, you can group each test for each attribute in the same module. For example
```
module('for @small', function () {
  test('if the value is true, ...)
  test('if the value is false, ...)
}
```

addon/components/o-s-s/progress-bar.ts
```diff
@@ -0,0 +1,50 @@
...
+  small?: boolean;
+}
+
+export default class OSSProgressBar extends Component<OSSProgressBarArgs> {
+  constructor(owner: unknown, args: OSSProgressBarArgs) {
+    super(owner, args);
+
+    assert(
+      '[component][OSS::ProgressBar] You must pass a numbered value between 0 and 100',
+      0 <= args.value && args.value <= 100
```
JulienVannier66, line 24:
```suggestion
      '[component][OSS::ProgressBar] You must pass a numbered value between 0 and 100',
      0 >= args.value && args.value <= 100
```
The range is not good here, negative value pass you range here

app/styles/atoms/progress-bar.less
OwenCoogan: Doesn't exist in DSM , checked with William & set the hex according to the existing color

tests/integration/components/o-s-s/progress-bar-test.ts
OwenCoogan: Didn't like the `Element | null `type in the assert , now I force it to be an `Element` only . If the element is null then the test fails anyway

tests/integration/components/o-s-s/progress-bar-test.ts
OwenCoogan: Yep that's done 👍

addon/components/o-s-s/progress-bar.ts
OwenCoogan: Sorted that one out too

addon/components/o-s-s/progress-bar.hbs
```diff
@@ -0,0 +1,19 @@
...
+  {{#if @label}}
+    <span class="oss-progress-bar--label">
+      {{@label}}
+    </span>
+  {{/if}}
+  <div class="oss-progress-bar">
+    <div 
+      class={{concat "oss-progress-bar--inner " this.computedStyles (if @small " oss-progress-bar--inner--small") }}
+      role="progressbar" aria-valuenow={{@value}}
+      aria-valuemin='0' aria-valuemax='100' style={{progressBarInnerStyle}}>
```
Miexil, line 11: single quotes

addon/components/o-s-s/progress-bar.hbs
```diff
@@ -0,0 +1,19 @@
+<div class="oss-progress-bar--wrapper">
+  {{#if @label}}
+    <span class="oss-progress-bar--label">
+      {{@label}}
+    </span>
+  {{/if}}
```
Miexil, line 6: Are we sure we want to handle the label in the component directly ? 🤔 
@phndiaye WDYT ?

addon/components/o-s-s/progress-bar.stories.js
```diff
@@ -0,0 +1,91 @@
...
+  small: true
+};
+
+const BasicUsageTemplate = (args) => ({
+  template: hbs`
+      <OSS::ProgressBar @value={{this.value}}
+                       @label={{this.label}}
+                       @valueIsVisible={{this.valueIsVisible}}
+                       @skin={{this.skin}}
+                       @small={{this.small}}  />
```
Miexil, line 85: Nit: wrap in a div with padding and black background so that we can see labels in storybook dark mode
see other stories files for examples (attributes on the top of my mind)

addon/components/o-s-s/progress-bar.stories.js
```diff
@@ -0,0 +1,91 @@
...
+      description: 'Displayed label',
+      table: {
+        type: {
+          summary: 'string'
+        },
+        defaultValue: { summary: undefined }
+      },
+      control: { type: 'text' }
+    },
+    valueIsVisible: {
```
Miexil, line 30: nit, rename to `displayValue`

addon/components/o-s-s/progress-bar.hbs
```diff
@@ -0,0 +1,19 @@
...
+  <div class="oss-progress-bar">
+    <div 
+      class={{concat "oss-progress-bar--inner " this.computedStyles (if @small " oss-progress-bar--inner--small") }}
+      role="progressbar" aria-valuenow={{@value}}
+      aria-valuemin='0' aria-valuemax='100' style={{progressBarInnerStyle}}>
+    </div>
+  </div>
+  {{#if @valueIsVisible}}
```
Miexil, line 14: nit, rename to `displayValue`

addon/components/o-s-s/progress-bar.ts
```diff
@@ -0,0 +1,43 @@
+import Component from '@glimmer/component';
+import { assert } from '@ember/debug';
+
+type ProgressBarSkins = 'primary' | 'warning' | 'success';
+
+interface OSSProgressBarArgs {
+  value: number;
+  label?: string;
+  valueIsVisible?: boolean;
```
Miexil, line 9: nit, rename to `displayValue`

addon/components/o-s-s/progress-bar.ts
```diff
@@ -0,0 +1,43 @@
...
+  skin?: ProgressBarSkins;
+  small?: boolean;
+}
+
+export default class OSSProgressBar extends Component<OSSProgressBarArgs> {
+  constructor(owner: unknown, args: OSSProgressBarArgs) {
+    super(owner, args);
+
+    assert('[component][OSS::ProgressBar] You must pass a numbered value between 0 and 100', args.value <= 100);
```
Miexil, line 18: Doesn't this raise an error if existence & type checking aren't done ?
maybe not but if it does, add `typeof args.value === 'number' && args.value >=0 && args.value <= 100`
(might as well check against negative values while we're there)

addon/components/o-s-s/progress-bar.ts
```diff
@@ -0,0 +1,43 @@
...
+    const stylesheet = document.styleSheets[0];
+    const keyframesRule = `@keyframes oss-progress-bar-animation {
+      from {
+        width: 0;
+      }
+      to {
+        width: ${this.args.value}%;
+      }
+    }`;
+    stylesheet.insertRule(keyframesRule, stylesheet.cssRules.length);
```
Miexil, line 29: Are we sure we want to do this like that ?
Can't we just define the keyframes in the less file and have it from 0 to 100 like the original upf-progress does ?

addon/components/o-s-s/progress-bar.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  get skin() {
```
Miexil, line 32: missing return type

addon/components/o-s-s/progress-bar.ts
```diff
@@ -0,0 +1,43 @@
...
+    return this.args.skin || 'primary';
+  }
+
+  get computedStyles() {
```
Miexil, line 36: missing return type

addon/components/o-s-s/progress-bar.ts
```diff
@@ -0,0 +1,43 @@
...
+    return 'oss-progress-bar--inner--' + this.skin;
+  }
+
+  get progressBarInnerStyle() {
```
Miexil, line 40: missing return type

app/styles/atoms/progress-bar.less
```diff
@@ -0,0 +1,41 @@
+.oss-progress-bar--wrapper {
+  min-width: 220px;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  p {
```
Miexil, line 6: I see no `<p>` tag in the template 🤔

app/styles/atoms/progress-bar.less
```diff
@@ -0,0 +1,41 @@
+.oss-progress-bar--wrapper {
+  min-width: 220px;
```
Miexil, line 2: Why define a min-width ? 🤔

app/styles/atoms/progress-bar.less
```diff
@@ -0,0 +1,41 @@
...
+    height: 10px;
+
+    &--small {
+      height: 4px;
+    }
+
+    &--primary {
+      background-color: var(--color-primary-500);
+    }
+    &--warning {
```
Miexil, line 33: missing line break

tests/integration/components/o-s-s/progress-bar-test.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  module('@value arg behaviour', function () {
+    test('It renders the progress bar inner with the correct width when @value is specified', async function (assert) {
+      await render(hbs`<OSS::ProgressBar @value={{this.checkedValue}} />`);
+
+      assert.dom('.oss-progress-bar').exists();
+      const innerBar = this.element.querySelector('.oss-progress-bar--inner') as Element;
+      const inlineStyle = innerBar.getAttribute('style');
+      assert.ok(inlineStyle?.includes(this.checkedValue), 'The inline style attribute is set to the proper value');
+    });
```
Miexil, line 36: This is the same test as line 19, no ?

tests/integration/components/o-s-s/progress-bar-test.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    test('It renders the progress bar inner with the correct width when @value is updated from one value to another', async function (assert) {
```
Miexil, line 38: nit: `the progress-bar--inner` or `the inner progress bar`

tests/integration/components/o-s-s/progress-bar-test.ts
```diff
@@ -0,0 +1,113 @@
...
+      await render(hbs`<OSS::ProgressBar @value={{this.checkedValue}} />`);
+
+      assert.dom('.oss-progress-bar').exists();
+      const innerBar = this.element.querySelector('.oss-progress-bar--inner') as Element;
+
+      let inlineStyle = innerBar.getAttribute('style');
+      assert.ok(inlineStyle?.includes(this.checkedValue), 'The inline style attribute is set to the proper value');
+      this.set('checkedValue', 100);
+      inlineStyle = innerBar.getAttribute('style');
+      assert.ok(inlineStyle?.includes('100'), 'The inline style attribute is set to the proper value');
```
Miexil, line 48: qUnit dom has a `hasStyle` check

```
assert.dom('.element').hasStyle({ 'border-left': '1px solid rgb(229, 231, 235)' });
```

can't this be used for this test ?

addon/components/o-s-s/progress-bar.ts
```diff
@@ -0,0 +1,43 @@
...
+import { assert } from '@ember/debug';
+
+type ProgressBarSkins = 'primary' | 'warning' | 'success';
+
+interface OSSProgressBarArgs {
+  value: number;
+  label?: string;
+  valueIsVisible?: boolean;
+  skin?: ProgressBarSkins;
+  small?: boolean;
```
Miexil, line 11: Was this asked by William ?

addon/components/o-s-s/progress-bar.stories.js
```diff
@@ -0,0 +1,91 @@
...
+      table: {
+        type: {
+          summary: 'boolean'
+        },
+        defaultValue: { summary: undefined }
+      },
+      control: { type: 'boolean' }
+    },
+    small: {
+      description: 'Boolean to adjust height of progress bar',
```
Miexil, line 41: Add
`If true, the height will be 4px. The default height is Xpx`

tests/integration/components/o-s-s/progress-bar-test.ts
Miexil: Also for the tests below for the `@small` param

addon/components/o-s-s/progress-bar.ts
OwenCoogan: I tried that & ended up not being able to override the 100% width attribute & had this in display : 

Capture d’écran 2023-11-24 à 14 43 44

app/styles/atoms/progress-bar.less
OwenCoogan: replaced with span

app/styles/atoms/progress-bar.less
OwenCoogan: Removed & will be managed from parent ( as seen in SB story )

addon/components/o-s-s/progress-bar.ts
OwenCoogan: The sync with william determined that they didn't want any real changes to it . The idea was only to mimic the one in prod in OSS . so I got that from what was specified in the suggested design on linear & what was done for upf-progress. @skin & @small are in use on the soft so they went in. @label & @labelIsVisible are in use in the sidepanel

addon/components/o-s-s/progress-bar.ts
JulienVannier66: Maybe with the CSS variable you can do something. You can update the CSS variable in JS and use it in the keyframe

phndiaye: +1 with handling this in w/ a CSS variable, or on the CSS side generally

app/styles/atoms/progress-bar.less
```diff
@@ -0,0 +1,42 @@
...
+  span {
+    font-weight: 200;
+    margin: 0 var(--spacing-px-6);
+  }
+}
+
+.oss-progress-bar {
+  width: 100%;
+  border-radius: var(--border-radius-sm);
+  background-color: #e5e5e5;
```
phndiaye, line 15: no matching variable here ?

app/styles/atoms/progress-bar.less
```diff
@@ -0,0 +1,42 @@
...
+
+    &--primary {
+      background-color: var(--color-primary-500);
+    }
+
+    &--warning {
+      background-color: var(--color-warning-500);
+    }
+
+    &--success {
```
phndiaye, line 38: why not put the `--modifier` on the parent class instead? (e.g: `.oss-progress-bar.oss-progress-bar--success`)

addon/components/o-s-s/progress-bar.ts
phndiaye: I think we should take the occasion to make this way simpler than it is currently then iterate instead of copying all of the current features.

For example, prefer a `size="sm"` arg instead of `small=true` arg here. I'm the one who did the old component and i clearly regretted that choice in some other component because you can end up doing something like: `large={{true}}` but nothing prevents you from doing `small={{true}} large={{true}}` together

addon/components/o-s-s/progress-bar.hbs
phndiaye, line 1: BEM wise, this is not really valid 😢 

With the rule being Block -> Element -> Modifier, this should be:

oss-progress-bar (serves as wrapper)-> oss-progress-bar__label
oss-progress-bar -> oss-progress-bar__value
.oss-progress-bar.oss-progress-bar--small -> oss-progress-bar__label (if small)

addon/components/o-s-s/progress-bar.hbs
```diff
@@ -0,0 +1,19 @@
+<div class='oss-progress-bar--wrapper'>
```
phndiaye, line 1: global comment on the file + add `...attributes` on this parent div

app/styles/atoms/progress-bar.less
OwenCoogan: Nope it wasn't in the dsm

Miexil: It might be something like Gray-100 or Gray-50

OwenCoogan: Added  --color-gray-200

app/styles/atoms/progress-bar.less
OwenCoogan: The inner takes the class because I don't want to override the one in the parent. Parent takes  --color-gray-200, inner takes modifier . I made it bem compliant though

addon/components/o-s-s/progress-bar.hbs
OwenCoogan: Changed , although the inner takes the size as it isn't displayed when no height is specified

OwenCoogan:
@phndiaye I added the size type to use "sm" or "lg" for now. Might be more but should be ok for now. I agree about iteration for sure , starting with replacing the label with a named block for the label that can take more than just a string.

addon/components/o-s-s/progress-bar.hbs
JulienVannier66: You don't update it yet, right?

addon/components/o-s-s/progress-bar.stories.js
```diff
@@ -0,0 +1,96 @@
...
+      },
+      control: { type: 'boolean' }
+    },
+    size: {
+      description: 'Adjust size',
+      table: {
+        type: {
+          summary: ProgressBarSizes.join('|')
+        },
+        defaultValue: { summary: 'sm' }
```
JulienVannier66, line 47: According to the code, the default value is `lg`

addon/components/o-s-s/progress-bar.hbs
```diff
@@ -0,0 +1,19 @@
...
+         role='progressbar' aria-valuenow={{@value}}
+         aria-valuemin='0' aria-valuemax='100' style={{progressBarWidthStyle}}>
+    </div>
+  </div>
+  {{#if @displayValue}}
+    <span class='oss-progress-bar__value'>
+      {{@value}}%
+    </span>
+  {{/if}}
+</div>
```
Miexil, line 19: Please use double quotes in template files

single quotes are okay inside double quotes, eg:
`class="button {{if @param 'other-class'}}'"`

addon/components/o-s-s/progress-bar.hbs
```diff
@@ -0,0 +1,19 @@
...
+    <span class='oss-progress-bar__label'>
+      {{@label}}
+    </span>
+  {{/if}}
+  <div class='oss-progress-bar__outer'>
+    <div
+         class={{concat 'oss-progress-bar__inner ' this.computedStyles}}
+         role='progressbar' aria-valuenow={{@value}}
+         aria-valuemin='0' aria-valuemax='100' style={{progressBarWidthStyle}}>
+    </div>
```
Miexil, line 12: Mis-alignment here.

addon/components/o-s-s/progress-bar.ts
```diff
@@ -0,0 +1,39 @@
...
+    assert(
+      '[component][OSS::ProgressBar] You must pass a numbered value between 0 and 100',
+      typeof args.value === 'number' && args.value >= 0 && args.value <= 100
+    );
+
+    document.documentElement.style.setProperty('--progress-bar-animation-width', `${args.value}%`);
+  }
+
+  get computedSize(): string {
+    return this.args.size ? 'oss-progress-bar__' + this.args.size : '';
```
phndiaye, line 29: so as said before it will be `oss-progress-bar--${size or skin}`

addon/components/o-s-s/progress-bar.hbs
OwenCoogan: Fixed in lastest commit

addon/components/o-s-s/progress-bar.hbs
OwenCoogan: Fixed in lastest commit

app/styles/molecules/progress-bar.less
```diff
@@ -0,0 +1,56 @@
+.oss-progress-bar {
+  display: flex;
+  width: 100%;
+  justify-content: space-between;
+  align-items: center;
+
+  &__sm {
```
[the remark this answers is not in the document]
OwenCoogan: Fixed in lastest commit

phndiaye (approved):

PR #346, opened 2023-12-05
Attribute-panel: Fix event fire on cancel & allow custom icon for edit button

aprentout:
### What does this PR do?

Related to: #[ENG-1569](https://linear.app/upfluence/issue/ENG-1569/[ember-influencer]-new-attributespanelmergefields-components-and-usage)

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/attributes-panel.hbs
+++ b/addon/components/o-s-s/attributes-panel.hbs
@@ -13,7 +13,7 @@
         {{#if (has-block "contextual-action")}}
           {{yield to="contextual-action"}}
         {{/if}}
-        <OSS::Button @icon="fa-pen" @square={{true}} {{on "click" this.toggleMode}}
+        <OSS::Button @icon={{this.editIcon}} @square={{true}} {{on "click" this.onEdit}}
                      data-control-name="attributes-panel-mode-switch-button" />
       </div>
     {{/if}}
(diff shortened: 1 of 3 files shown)


Miexil (approved):

addon/components/o-s-s/attributes-panel.hbs
```diff
@@ -13,7 +13,7 @@
         {{#if (has-block "contextual-action")}}
           {{yield to="contextual-action"}}
         {{/if}}
-        <OSS::Button @icon="fa-pen" @square={{true}} {{on "click" this.toggleMode}}
+        <OSS::Button @icon={{this.editIcon}} @square={{true}} {{on "click" this.onEdit}}
```
phndiaye, line 16: can you also precise the `size` ? we did something a while ago that makes square buttons small by default but that's not really explicit

addon/components/o-s-s/attributes-panel.ts
```diff
@@ -25,9 +26,18 @@ export default class OSSAttributesPanel extends Component<OSSAttributesPanelArgs
...
+    return this.args.customEditIcon ?? 'fa-pen';
+  }
+
   @action
   toggleMode(): void {
     this.modeSelected = this.modeSelected === 'view' ? 'edition' : 'view';
+  }
+
+  @action
+  onEdit(): void {
```
phndiaye, line 39: what about an event.stopPropagation even though it's not really useful now ? but if one day `args.onEdit` opens a modal, we will be happy to have it ;)

phndiaye (approved):

PR #349, opened 2023-12-19
Added: OSS::ModalDialog @enqueue parameter + tests & doc

Miexil:
### What does this PR do?
Adds an `enqueue` parameter to OSS::ModalDialog which prevents the current modal from opening if other oss-modal-dialogs are already opened. The modal finally opens when other instances are closed.

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled


addon/components/o-s-s/modal-dialog.ts
```diff
@@ -1,20 +1,29 @@
 import { assert } from '@ember/debug';
 import Component from '@glimmer/component';
 import BaseModal, { BaseModalArgs } from './private/base-modal';
+import { tracked } from '@glimmer/tracking';
```
phndiaye, line 4: nit: can you swap these two 😛

phndiaye (approved):

JulienVannier66 (approved):

PR #350, opened 2023-12-26
[VEL-1590] Road to Glint support: Stage 1 — Update core dependencies

phndiaye:
### What does this PR do?

On the road to add support for Glint throughout this package, this PR starts by updating some of the core dependencies:

- ember-auto-import
- ember-cli-typescript
- ember-cli-typescript-blueprints: point to a branch fixing the generated files when using `ember g component`

The addon now needs Node v16 to run locally.
Tested against both an engine and an Ember app and it works like a charm.

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled


OwenCoogan (commented):
G2G for me but would need all sets of eyes on this one I think

.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-  NODE_VERSION: '14'
+  NODE_VERSION: '16'
```
Miexil, line 13: does this mean we are no longer using `14.18.2` locally ?

addon/components/o-s-s/attribute/country.ts
```diff
@@ -1,5 +1,5 @@
 import Component from '@glimmer/component';
-import { CountryData, countries } from '@upfluence/oss-components/utils/country-codes';
+import { type CountryData, countries } from '@upfluence/oss-components/utils/country-codes';
```
Miexil, line 2: 🤔

addon/components/o-s-s/button-dropdown.ts
```diff
@@ -57,14 +57,14 @@ export default class extends Component<OSSButtonDropdownArgs> {
 
   @action
   setupChildrenClickHandler(element: HTMLElement): void {
-    element.querySelectorAll('.oss-button-dropdown__item').forEach((child: HTMLElement) => {
+    element.querySelectorAll('.oss-button-dropdown__item').forEach((child: Element) => {
```
Miexil, line 60: isn't `HTMLElement` a more precise `Element`? Aren't we losing precision with this change ?

.nvmrc
```diff
@@ -0,0 +1 @@
+v16.19.0
```
Miexil, line 1: hmmmm
I remember having issues with some projects with this node version. Why are we moving to this one for this project ?

addon/components/o-s-s/password-input.ts
```diff
@@ -130,7 +130,10 @@ export default class OSSPasswordInput extends Component<OSSPasswordInputArgs> {
   }
 
   private testAllValidators(): boolean {
-    return this.inputValidators.every((index: InputValidator) => this.validatorSet[index].regex.test(this.args.value!));
+    return this.inputValidators.every((key: InputValidator) => {
+      if (!this.validatorSet[key]) return false;
```
Miexil, line 134: just for extra safety, right ?

addon/components/o-s-s/phone-number-input.ts
```diff
@@ -41,7 +41,7 @@ export default class OSSPhoneNumberInput extends Component<OSSPhoneNumberInputAr
       typeof this.args.onChange === 'function'
     );
 
-    this.selectedCountry = this._countries[0];
+    this.selectedCountry = this._countries[0]!;
```
Miexil, line 44: Is this throwing a warning on the new node version ? that's weird 🤔

addon/components/o-s-s/social-post-badge.stories.js
```diff
@@ -78,7 +78,7 @@ const defaultArgs = {
 const Template = (args) => ({
   template: hbs`
     <OSS::SocialPostBadge @postType={{this.postType}} @onToggle={{this.onToggle}} @selected={{this.selected}}
-                           @plain={{this.plain}}/>        
+                           @plain={{this.plain}}/>
```
Miexil, line 81: may as well fix alignment

package.json
```diff
@@ -31,75 +31,78 @@
...
+    "@types/rsvp": "^4.0.8",
     "@types/sinon": "^10.0.2",
     "@typescript-eslint/parser": "^5.0.0",
     "babel-eslint": "^10.1.0",
     "babel-loader": "^8.1.0",
-    "ember-auto-import": "^1.5.3",
     "ember-cli": "~3.28.6",
     "ember-cli-inject-live-reload": "^2.1.0",
-    "ember-cli-typescript-blueprints": "^3.0.0",
+    "ember-cli-typescript-blueprints": "muziejus/ember-cli-typescript-blueprints#ember-4-update",
```
Miexil, line 103: are we sure about this?

.github/workflows/ci.yml
phndiaye: Considering we are not actively using Node code ourselves (beside the custom linter rules) and that ember-cli is compatible there should not be any issues.

But i remember having issues too just not in this project, I guess we'll see them down the road while migrating to pnpm (which requires Node >= 16)

In any case, this doesn't force us to migrate everything till we move to pnpm. (IMO we should go even further and move to Node 18 but that's too much a change and one thing a time 😅 )

package.json
phndiaye: Forked and used our version for extra safety

addon/components/o-s-s/attribute/country.ts
phndiaye: This is a result of https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax being enabled by default in TS5. 
IMO it's a good thing but as we were not using it before, it leads to a lot of files changed. But it can be disabled too

.nvmrc
phndiaye: My bad, replied on the wrong comment: https://github.com/upfluence/oss-components/pull/350#discussion_r1461815512.

addon/components/o-s-s/password-input.ts
phndiaye: yep! but i got an error about it the first time i did the migration and now it works without this change 😩

addon/components/o-s-s/phone-number-input.ts
phndiaye: More like the TS version not letting it slide anymore. I think they fixed something because it was never supposed to let it slide IMO

package.json
```diff
@@ -116,9 +119,10 @@
     "loader.js": "^4.7.0",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.5.1",
-    "qunit": "^2.17.2",
-    "qunit-dom": "^1.6.0",
-    "typescript": "^4.4.4"
+    "qunit": "^2.20.0",
```
JulienVannier66, line 122: `qunit` is in the `peerDependencies`, do we keep here too?

JulienVannier66 (approved):
Otherwise LGTM

package.json
phndiaye: Yep! yarn does not auto-install peers (pnpm does 🙈 ) so have to list it in both keys

Miexil (approved):

PR #351, opened 2023-12-26
[VEL-1590]  Road to Glint support: Stage 2 — Kickoff Glint support

phndiaye:
### Background

Although we have pretty good Typescript support across the component library, it unfortunately can't see what's happening in templates, meaning we miss out on any typechecking and other helpful annotations it can provide.

[Glint](https://typed-ember.gitbook.io/glint) enables TypeScript to understand your templates as well, meaning you get end-to-end type safety for your entire application. You also get editor support for things like documentation and type info on hover, jumping to definitions, and automated codefixes and refactorings.

Interesting take on this too:

- https://youtu.be/8rnmGGY5rhk?t=1079

### What does this PR do?

Built on top of the core dependencies updates, this PR introduces Glint setup and:

- Updates components' signatures to the new format supported by Glint (via `ember-codemod-args-to-signature` codemod) 
- marks most components as type unsafe for the moment so we can gradually update
- Make 2 helpers (fa-icon-style & fa-icon-value) type safe to serve as examples
- Get helper typings for ember-intl's `t` helper and ember-truth-helpers common helpers (via https://github.com/Gavant/glint-template-types) as well as modifier typings for ember-render-modifiers (`did-insert`, `will-destroy`, etc.)

For the moment, only a couple component are considered "type-safe" and everything else is marked as type-unsafe. To enable full support, the best way (i believe) would be to first port components that depends on nothing (just HTML, no helpers, not using other components) then go up the chain with "composite" components (molecules?) and finish w/ the most complex ones.

To enable editor support for Glint, one needs to install the dedicated Glint Template Server extension:

- VSCode: https://marketplace.visualstudio.com/items?itemName=typed-ember.glint-vscode
- Neovim: via native LSP support (nvim_lspconfig + Mason is good pair here)

### What will affect us?

You'll notice that there are `glint-nocheck` comments in both the templates and the tests. This is what allows us to mark the components as type-unsafe so we can incrementally migrate them later.

While it doesn't make the templates way uglier, it does for tests files until we make them typesafe. Eventually, most of those comments will disappear as we increase the typesafety coverage.

However, It also forces us to pass the `TestContext` when rendering a component w/ a `this` usage in the args. This is something that's a good practice anyway but we were not doing it until now and it can affect the DX a bit. IMO, this can motivate us to always define a global `renderComponent` function in the tests so we don't have to put that `render<TestContext>(...args)` everytime.

Unfortunately (or fortunately 🤷‍♂️ ), Glint covers the same file covered by the `tsconfig` so we cannot opt-out of it on `tests` directory.

Final note: this PR only sets kicks off Glint support inside this package. Other ones in apps will follow to see the types are used.

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled


tests/integration/helpers/redirect-to-test.ts
```diff
@@ -13,15 +13,17 @@ module('Integration | Helper | redirect-to', function (hooks) {
...
+    await render<TestContext>(hbs`<div role="button" {{on "click" (redirect-to url=this.url)}}>link</div>`);
     await click('div');
     assert.ok(this.redirectStub.calledOnceWithExactly(this.url, '_self'));
     sinon.restore();
   });
 
   test('it redirects to the url on the provided target', async function (assert) {
     this.redirectStub = sinon.stub(window, 'open');
-    await render(hbs`<div {{on "click" (redirect-to url=this.url target="_blank")}}>link</div>`);
+    await render<TestContext>(
```
phndiaye, line 24: `TestContext` usage example.

phndiaye:
Closing since this did not have any advancement since it was tried. Will serve as inspiration if we want to explore the topic in the future.

PR #366, opened 2024-02-28
use floating-ui/dom to manage OSS::Select & OSS::PowerSelect dropdowns

phndiaye:
### What does this PR do?

Related to: #

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled


addon/components/o-s-s/select.ts
```diff
@@ -23,6 +27,12 @@ export default class OSSSelect extends Component<OSSSelectArgs> {
 
   @tracked displaySelect: boolean = false;
 
+  cleanupDrodpownAutoplacement?: () => void;
```
JulienVannier66, line 30: Why added a `?` here?

addon/components/o-s-s/select.hbs
```diff
@@ -17,37 +23,45 @@
...
+      <OSS::InfiniteSelect
+        @items={{@items}}
+        @onSearch={{this.onSearch}}
+        @onSelect={{this.onSelect}}
+        @searchEnabled={{this.searchEnabled}}
+        @searchPlaceholder={{this.searchPlaceholder}}
+        id={{this.portalId}}
+        class="margin-top-px-0"
+        {{on-click-outside this.onClickOutside}}
+        {{on "click" this.noop}}
```
JulienVannier66, line 37: Do you really need this one?

addon/components/o-s-s/select.ts
```diff
@@ -66,6 +76,10 @@ export default class OSSSelect extends Component<OSSSelectArgs> {
     return classes.join(' ');
   }
 
+  noop(event: Event): void {
+    event.stopPropagation();
+  }
```
JulienVannier66, line 81: Same here

tests/integration/components/o-s-s/select-test.ts
```diff
@@ -31,7 +31,7 @@ module('Integration | Component | o-s-s/select', function (hooks) {
       );
 
       await click('.upf-input div');
-      await click('.upf-infinite-select .upf-infinite-select__item:last-child');
+      await click(document.querySelector('.upf-infinite-select .upf-infinite-select__item:last-child')!);
```
JulienVannier66, line 34: Why using `querySelector` now?

addon/components/o-s-s/select.ts
phndiaye: because it doesn't have a value from the start

tests/integration/components/o-s-s/select-test.ts
phndiaye: previously, it was because the dropdown was being portaled into the body, but that's not needed anymore

addon/components/o-s-s/select.ts
phndiaye: it's because when the dropdown is portaled into the body, clicking in it (not the options but like empty state or corners) actually counts as a click outside 😅  this catches it

addon/components/o-s-s/select.hbs
phndiaye: replied on the https://github.com/upfluence/oss-components/pull/366#discussion_r1514798546

JulienVannier66 (approved):

addon/components/o-s-s/power-select.hbs
Miexil, line 1: All quotes have been transformed to single quotes ⚠️ :)

addon/components/o-s-s/power-select.ts
```diff
@@ -38,19 +50,47 @@ export default class OSSPowerSelect extends Component<OSSPowerSelectArgs> {
   }
 
   @action
-  toggleSelect(event: MouseEvent): void {
-    event.stopPropagation();
-    this.displaySelect = !this.displaySelect;
+  handleSelectorClose(event: ToggleEvent & { target: HTMLDetailsElement }) {
```
Miexil, line 53: : void

addon/components/o-s-s/power-select.ts
```diff
@@ -38,19 +50,47 @@ export default class OSSPowerSelect extends Component<OSSPowerSelectArgs> {
...
+    super.onClickOutside(_, event);
+    this.cleanupDrodpownAutoplacement?.();
     this.args.onSearch?.('');
+    document.querySelector(`#${this.portalId}`)?.remove();
+  }
+
+  @action
+  registerContainer(element: HTMLDetailsElement): void {
+    super.registerContainer(element);
+    this.portalTarget = isTesting() ? this.container : document.body;
```
Miexil, line 94: why not leave it in the body even when testing ? 🤔

Miexil: messes up with the assert.dom method ?
I'm sure we can still document.queryElementBy* though 🤷

Miexil (approved):

Miexil (approved):

PR #372, opened 2024-03-29
Add support for OSS::AvatarGroup component

phndiaye:
### What does this PR do?

- Add support for OSS::AvatarGroup component
- Add support for loading/placeholder states on OSS::Avatar

Screenshot 2024-03-29 at 14 54 51
Screenshot 2024-03-29 at 14 54 38

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/avatar-group.stories.js
+++ b/addon/components/o-s-s/avatar-group.stories.js
@@ -1,0 +1,92 @@
+import { hbs } from 'ember-cli-htmlbars';
+
+const SizeTypes = ['xs', 'sm', 'md', 'lg'];
+
+export default {
+  title: 'Components/OSS::AvatarGroup',
+  argTypes: {
+    avatars: {
+      description: 'Avatars to display',
+      table: {
+        type: {
+          summary: '{ image?: string, initials?: string }[]'
+        }
+      },
+      control: {
+        type: '{ image?: string, initials?: string }[]'
+      }
+    },
+    size: {
+      description: 'Adjust the size of the avatars',
+      table: {
+        type: {
+          summary: SizeTypes.join('|')
+        },
+        defaultValue: { summary: 'md' }
+      },
+      options: SizeTypes,
+      control: { type: 'select' }
+    },
+    max: {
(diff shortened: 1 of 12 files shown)


addon/components/o-s-s/avatar.ts
```diff
@@ -16,6 +16,8 @@ interface OSSAvatarArgs {
   image?: string;
   initials?: string;
   size?: SizeType;
+  loading?: boolean;
+  isPlaceholder?: boolean;
```
JulienVannier66, line 20: Is it really a placeholder here? When it is true, you add a class with a background color

app/styles/molecules/avatar-group.less
```diff
@@ -0,0 +1,15 @@
...
+
+  & > .upf-avatar {
+    background-color: var(--color-white);
+    border: 1px solid var(--color-white);
+    border-radius: 50%;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+
+    margin-right: -6px;
```
JulienVannier66, line 12: css variable?

addon/components/o-s-s/avatar.ts
phndiaye: don't have any good name 😅  ahah

addon/components/o-s-s/avatar-group.stories.js
```diff
@@ -0,0 +1,92 @@
...
+        type: {
+          summary: 'number'
+        },
+        defaultValue: { summary: 'undefined' }
+      },
+      control: {
+        type: 'number'
+      }
+    }
+  }
```
Miexil, line 64: maybe add a param object with description as we have for other components :
```
parameters: {
    docs: {
      description: {
        component:
          'A component to automaticaly wrap link into an anchor element or a LinkTo helper based on application router knowledge.'
      },
      iframeHeight: 100
    }
  }
```

Miexil (approved):

PR #398, opened 2024-06-25
OSS::Copy: render nothing if the clipboard-write permission isn't granted

phndiaye:
### What does this PR do?

OSS::Copy: render nothing if the clipboard-write permission isn't granted

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/copy.ts
+++ b/addon/components/o-s-s/copy.ts
@@ -4,6 +4,7 @@ import { tracked } from '@glimmer/tracking';
 import { inject as service } from '@ember/service';
 
 import ToastService from '@upfluence/oss-components/services/toast';
+import type { IntlService } from 'ember-intl';
 
 interface OSSCopyArgs {
   value: string;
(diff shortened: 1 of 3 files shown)


addon/components/o-s-s/copy.ts
```diff
@@ -4,17 +4,31 @@ import { tracked } from '@glimmer/tracking';
...
 
+  constructor(owner: unknown, args: OSSCopyArgs) {
+    super(owner, args);
+
+    navigator.permissions
+      .query({ name: 'clipboard-write' as PermissionName })
+      .then(({ state }) => {
+        this.accessibleClipboard = state === 'granted';
+      })
+      .catch(() => {});
```
Miexil, line 29: any reason to declare catch if we're not using it for anything ?

Miexil (approved):

addon/components/o-s-s/copy.ts
phndiaye: it's so that if the feature detection somehow fail, we consider that it's not accessible but don't want to send it to Sentry either. Or should we? 🤔

phndiaye: here are the exceptions that could be thrown https://developer.mozilla.org/en-US/docs/Web/API/Permissions/query#exceptions

PR #410, opened 2024-08-05
Get rid of legacy radio-button component

phndiaye:
### What does this PR do?

Get rid of legacy radio-button component as all occurences has been replaced by either `OSS::RadioButton` or `OSS::ToggleButtons` depending on the context.

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled


app/styles/shims/_all.less
```diff
@@ -79,7 +79,6 @@
 // These are the colors defined by the Upfluence Visual Design Guidelines.
 // --------------------------------------------------
 
-@upf-primary-bright-purple: #0d0de6;
```
phndiaye, line 82: Looks like it has all been replaced w/ `color-primary-500` 🥳 

https://github.com/search?q=org%3Aupfluence+upf-primary-bright-purple++NOT+is%3Aarchived&type=code

nathalieArnoux (approved):

JulienVannier66 (approved):

PR #419, opened 2024-08-30
Added input autofocus modifier

nathalieArnoux:
### What does this PR do?

Related to: #[VEL-3542](https://linear.app/upfluence/issue/VEL-3542/create-autofocus-modifier)

https://www.loom.com/share/ea9dd0960ade47d99ec0df63e5c00984?sid=74e0cee2-e512-4656-94dc-daaaf3322b6e

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/modifiers/Modifiers.stories.mdx
+++ b/addon/modifiers/Modifiers.stories.mdx
@@ -11,19 +11,19 @@ import { Meta } from '@storybook/addon-docs/blocks';
 # Modifiers
 
 According to Ember JS:
+
 <figure>
   <blockquote cite="https://www.huxley.net/bnw/four.html">
-    <p>
-      Modifiers are a basic primitive for interacting with the DOM in Ember.
-    </p>
+    <p>Modifiers are a basic primitive for interacting with the DOM in Ember.</p>
   </blockquote>
   <figcaption>— Ember JS</figcaption>
 </figure>
 
-| Modifier         | Description                                                                  | Usage                                                                          | Default                                                                                                   |
-|------------------|------------------------------------------------------------------------------|--------------------------------------------------------------------------------| ----------------------------------------------------------------------------------------------------------|
-| enable-popover   | Adds a Bootstrap's Popover to the element                                    | `<button {{enable-popover [trigger]}}>Popover</button>`                        | Default Trigger : **hover**                                                                               |
-| enable-dropdown  | Toggles Bootstrap's dropdown on an element                                   | `<a {{enable-dropdown}}>Dropdown Trigger link/button</a>`                      |                                                                                                           |
-| on-click-outside | Triggers the passed callback function when a user clicks outside the element | `<div {{on-click-outside callbackFunction}}>Floating Div</div>`                |                                                                                                           |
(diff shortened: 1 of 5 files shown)


OwenCoogan (approved):

addon/modifiers/enable-input-autofocus.ts
```diff
@@ -0,0 +1,35 @@
...
+      if (!localElement.disabled) {
+        localElement.scrollIntoView({ behavior: 'smooth', block: 'center' });
+        localElement.focus({ preventScroll: true });
+      } else {
+        return;
+      }
+    },
+
+    destroyModifier() {
+      // We don't need to do anything here, but a function
```
aprentout, line 30: Nit: Don't know what others will think, but are those comments required ?

aprentout (approved):

Xegushu (approved):

addon/modifiers/enable-input-autofocus.ts
OwenCoogan: Well we don't need 2 lines to say that it needs to remain blank but I'm not too keen on removing it either , but up to you 👍

addon/modifiers/Modifiers.stories.mdx
```diff
@@ -11,19 +11,19 @@ import { Meta } from '@storybook/addon-docs/blocks';
...
-| scroll-shadow    | Add bottom shadow for scrollable element                                     | `<div {{scroll-shadow color="field"}}>`                                        | color: "default" (white)                                                                                           |
+| Modifier               | Description                                                                                | Usage                                                           | Default                     |
+| ---------------------- | ------------------------------------------------------------------------------------------ | --------------------------------------------------------------- | --------------------------- |
+| enable-popover         | Adds a Bootstrap's Popover to the element                                                  | `<button {{enable-popover [trigger]}}>Popover</button>`         | Default Trigger : **hover** |
+| enable-input-autofocus | Focuses on selected input with smooth scroll                                               | `<input type="text" {{enable-input-autofocus}}/>`               |                             |
```
Miexil, line 25: the* selected input 
but also, that's not completely true because you've handled the case where we can add the modifier on a parent element and the modifier will still target the first available input;

addon/modifiers/enable-input-autofocus.stories.js
```diff
@@ -0,0 +1,22 @@
+import { hbs } from 'ember-cli-htmlbars';
+
+export default {
+  title: 'Helpers & Modifiers/Modifiers/Input-Autofocus/Definition',
+  parameters: {
+    docs: {
+      description: {
+        component: 'A modifier to autofocus on selected input or first non-disabled input child.'
```
Miexil, line 8: on the* selected input or the* first non-disabled ...

addon/modifiers/enable-input-autofocus.ts
```diff
@@ -0,0 +1,35 @@
+// @ts-ignore
+import { setModifierManager, capabilities } from '@ember/modifier';
+
+export default setModifierManager(
+  () => ({
+    capabilities: capabilities('3.22'),
+
+    createModifier() {
```
Miexil, line 8: no method has a return value, I'm guessing this is also the case in other modifiers.
Maybe add some ?

PR #427, opened 2024-09-17
na/DRA-1046 - Added: safe string acceptance on oss-dialog title

nathalieArnoux:
### What does this PR do?

Added the option to pass a safe string to the title of the OSS::Dialog component.

Necessary for: #[DRA-1046](https://linear.app/upfluence/issue/DRA-1046/dialog-box-to-confirm-if-users-wants-to-define-default-account-email)

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/dialog.ts
+++ b/addon/components/o-s-s/dialog.ts
@@ -1,6 +1,7 @@
 import BaseModal, { type BaseModalArgs } from './private/base-modal';
 import { action } from '@ember/object';
 import { assert } from '@ember/debug';
+import { isEmpty } from '@ember/utils';
 
 type Skin = 'alert' | 'primary' | 'error';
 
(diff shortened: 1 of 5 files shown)


addon/components/o-s-s/dialog.ts
```diff
@@ -46,4 +50,8 @@ export default class OSSDialog extends BaseModal<OSSDialogArgs> {
   onInit(elem: HTMLElement): void {
     this.initialize(elem, false);
   }
+
+  private isSafeString(data: any): boolean {
+    return typeof data === 'object' && !isEmpty(data.string);
+  }
```
JulienVannier66, line 56: According to the `ember-intl` module, you can maybe use `import type { SafeString } from '@ember/template';` for the typing

Xegushu (approved):

addon/components/o-s-s/dialog.ts
aprentout: @nathalieArnoux 🙏 As Julien write, ember-template already has type & validator, it could be great to use them instead
https://api.emberjs.com/ember/5.11/functions/@ember%2Ftemplate/isHTMLSafe

tests/dummy/app/templates/application.hbs
```diff
@@ -1051,7 +1051,7 @@
 
 {{#if this.showDialog}}
   <OSS::Dialog
-    @title="You are about to discard your changes"
+    @title={{t "oss-components.dialog.title" htmlSafe=true}}
```
OwenCoogan, line 1054: Could we avoid setting HtmlSafe to true by defaut ?

JulienVannier66 (approved):

OwenCoogan (approved):

phndiaye (approved):

PR #430, opened 2024-09-27
Added square arg to OSS::ButtonDropdown

OwenCoogan:
### What does this PR do?

Capture d’écran 2024-09-27 à 16 39 38

https://www.loom.com/share/079b104057cb4022a92a04fdfa4e2055

Related to: #

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/button-dropdown.stories.js
+++ b/addon/components/o-s-s/button-dropdown.stories.js
@@ -25,6 +25,16 @@ export default {
       },
       control: { type: 'boolean' }
     },
+    hideArrow: {
+      description: 'Optional - An argument that displays the dropdown arrow icon',
+      table: {
+        type: {
+          summary: 'boolean'
+        },
+        defaultValue: { summary: 'false' }
+      },
+      control: { type: 'boolean' }
+    },
     label: {
       description: 'The label of the button label',
       table: {
@@ -60,6 +70,7 @@ const defaultArgs = {
   icon: 'far fa-user',
   label: 'Actions',
   square: false,
+  hideArrow: false,
   mainAction: null
 };
 
@@ -67,12 +78,13 @@ const withMainActionArgs = {
   icon: 'far fa-user',
   label: 'Actions',
   square: false,
+  hideArrow: false,
   mainAction: action('mainAction')
 };
 
 const Template = (args) => ({
   template: hbs`
-    <OSS::ButtonDropdown @icon={{this.icon}} @label={{this.label}} @mainAction={{this.mainAction}} @square={{this.square}}>
+    <OSS::ButtonDropdown @icon={{this.icon}} @label={{this.label}} @mainAction={{this.mainAction}} @square={{this.square}} @hideArrow={{this.hideArrow}}>
       <:items>
         <div class="oss-button-dropdown__item font-color-gray-900">
           <OSS::Icon @style="solid" @icon="fa-share" /> Share
--- a/addon/components/o-s-s/button-dropdown.stories.js
+++ b/addon/components/o-s-s/button-dropdown.stories.js
@@ -15,6 +15,16 @@ export default {
       },
       control: { type: 'text' }
     },
+    square: {
+      description: 'An arg that makes the button shape squared',
+      table: {
+        type: {
+          summary: 'boolean'
+        },
+        defaultValue: { summary: 'false' }
+      },
+      control: { type: 'boolean' }
+    },
     label: {
       description: 'The label of the button label',
       table: {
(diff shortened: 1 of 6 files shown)


addon/components/o-s-s/button-dropdown.stories.js
```diff
@@ -15,6 +15,16 @@ export default {
       },
       control: { type: 'text' }
     },
+    square: {
+      description: 'An arg that makes the button shape squared',
```
Miexil, line 19: nit: argument :)

Miexil (commented):
It's not clear how this displays from your screenshot, do you mind sharing another picture from the localhost rendering ? 😇

addon/components/o-s-s/button-dropdown.stories.js
OwenCoogan: Updated

addon/components/o-s-s/button-dropdown.stories.js
```diff
@@ -49,20 +59,22 @@ export default {
 const defaultArgs = {
   icon: 'far fa-user',
   label: 'Actions',
+  square: false,
```
nathalieArnoux, line 62: Is it necessary to set "square" as false here since it's an optional argument?

addon/components/o-s-s/button-dropdown.stories.js
```diff
@@ -49,20 +59,22 @@ export default {
...
   mainAction: null
 };
 
 const withMainActionArgs = {
   icon: 'far fa-user',
   label: 'Actions',
+  square: false,
```
nathalieArnoux, line 69: Same question here

addon/components/o-s-s/button-dropdown.stories.js
```diff
@@ -15,6 +15,16 @@ export default {
       },
       control: { type: 'text' }
     },
+    square: {
+      description: 'An argument that makes the button shape squared',
```
nathalieArnoux, line 19: Should we mention this argument is optional?

nathalieArnoux (approved):

addon/components/o-s-s/button-dropdown.stories.js
Miexil: necessary, no, but it's fine as is.

tests/integration/components/o-s-s/button-dropdown-test.ts
```diff
@@ -23,6 +23,24 @@ module('Integration | Component | o-s-s/button-dropdown', function (hooks) {
...
 
+  test('the button is displayed as square when @square is provided', async function (assert) {
+    await render(
+      hbs`<OSS::ButtonDropdown @icon="far fa-users" @square={{true}}><:items><div class="oss-button-dropdown__item">foo</div></:items></OSS::ButtonDropdown>`
+    );
+
+    assert.dom('.oss-button-dropdown__trigger-square').exists();
+  });
+
+  test('the button is hidden when @hideArrow is provided', async function (assert) {
```
Miexil, line 34: this label is wrong, I believe you wanted to write the arrow / caret

Miexil (commented):
I initially thought this dropdown button looked like the one we had in the composer :
Screenshot 2024-10-01 at 08 44 48

but it turns out there is no seperation between the text and the arrow/caret in the one from OSS-components :/ 

I asked for the hideArrow as an option, thinking it was the same component as in the composer, but it does not look good right now to have a square button with an arrow displayed.
Current implementation :
Screenshot 2024-10-01 at 08 49 44

I think in the case of rounded + display arrow, we should change the `justify-content` to `center` and add a flex gap of 3px between the fa-icon and the arrow. It would look like this :

Screenshot 2024-10-01 at 08 49 32

but we need to be careful to keep the OG layout untouched

addon/components/o-s-s/button.ts
```diff
@@ -78,6 +78,7 @@ interface ButtonArgs {
   label?: string;
   theme?: string;
   square?: boolean;
+  displayArrow?: boolean;
```
aprentout, line 81: is this really used ?

OwenCoogan: That was a whoopsie ^^' Removed it

tests/integration/components/o-s-s/button-dropdown-test.ts
OwenCoogan: Changed label

Miexil (approved):

Xegushu (approved):

JulienVannier66 (approved):

PR #440, opened 2024-10-24
Column filter input field issue on Firefox

Xegushu:
### What does this PR do?
This PR removes the default spinner buttons on the slider in Firefox

image

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/app/styles/base/_slider.less
+++ b/app/styles/base/_slider.less
@@ -34,6 +34,11 @@
       margin: 0;
     }
 
+    input[type='number'] {
+      -moz-appearance: textfield;
+      margin: 0;
+    }
+
     &--disabled {
       .oss-slider {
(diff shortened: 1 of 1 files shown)


OwenCoogan (approved):

aprentout (approved):

app/styles/base/_slider.less
aprentout, line 1: To be very very very nitty, `-moz-appearance: none;` line 33 is useless as firefox doesn't handle `::-webkit-inner-spin-button` or `::-webkit-outer-spin-button` so we could probably remove it

JulienVannier66 (approved):

PR #445, opened 2024-11-05
Removed: bootstrap's col-xs usage from InputWrapper

Miexil:
### What does this PR do?
Removes the `col-xs` bootstrap class from the soon to be deprecated `InputWrapper` component

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/app/styles/base/_inputs.less
+++ b/app/styles/base/_inputs.less
@@ -192,7 +192,6 @@ label {
 
   .upf-input-feedback {
     position: relative;
-    min-height: 1px;
     padding-right: var(--spacing-px-12);
     padding-left: var(--spacing-px-12);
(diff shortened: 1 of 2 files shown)


OwenCoogan (approved):

phndiaye (approved):

app/styles/base/_inputs.less
```diff
@@ -191,6 +191,13 @@ label {
   position: relative;
 
   .upf-input-feedback {
+    position: relative;
+    min-height: 1px;
```
aprentout, line 195: Is the min-height useful ?

PR #446, opened 2024-11-06
LIB-REMOVAL: removed upf-slider & ion rangeslider lib

OwenCoogan:
### What does this PR do?

Related to: #

### Good PR checklist
- [ ] Title makes sense
- [ ] Is against the correct branch
- [ ] Properly assigned
- [ ] Added/updated tests
- [ ] Properly labeled


nathalieArnoux (approved):

Miexil (approved):

phndiaye (approved):

addon/components/o-s-s/slider.stories.js
```diff
@@ -78,6 +78,16 @@ export default {
         defaultValue: { summary: '50' }
       }
     },
+
+    tooltiplabel: {
+      control: 'text',
+      description: 'Specific tooltip label for the slider handle',
+      defaultValue: `Value : ${defaultArgs.defaultValue}%, or apples if you want`,
```
Miexil, line 85: this is misleading ^^
should only be `defaultValue: ${defaultArgs.defaultValue}%`

addon/components/o-s-s/slider.ts
```diff
@@ -72,8 +73,19 @@ export default class SliderComponent extends Component<SliderComponentArgs> {
...
+
+    if (value > max) {
+      (event.target as HTMLInputElement).value = String(max);
+      this.args.onChange(max);
+    } else if (value < min) {
+      (event.target as HTMLInputElement).value = String(min);
+      this.args.onChange(min);
+    } else {
+      this.args.onChange(value);
+    }
```
Miexil, line 88: Does this not break existing usages of `min / max` in other components ?
Not sure I understand the reasoning behind these changes 🤔

app/styles/base/_slider.less
```diff
@@ -4,9 +4,11 @@
 
     .oss-slider__tooltip {
       text-align: center;
+      white-space: nowrap;
       left: var(--tooltip-left-position);
       top: 30px;
       transition: opacity 300ms;
+      z-index: 1000;
```
Miexil, line 11: I think this was one of the attempts to have it display over the tether-element in the crm context.
Don't think it's required anymore ?

tests/integration/components/o-s-s/slider-test.ts
```diff
@@ -38,6 +38,15 @@ module('Integration | Component | o-s-s/slider', function (hooks) {
     assert.dom('.oss-slider__number-input input').isDisabled();
   });
 
+  test('it renders with a specific tooltip label when specifiec', async function (assert) {
```
Miexil, line 41: when specified*

addon/components/o-s-s/slider.stories.js
OwenCoogan: Cleaned up

addon/components/o-s-s/slider.ts
OwenCoogan: Retested a bunch where we use it on the soft & played around with it on storybook, all good

This is what I managed to cap max & min as html range inputs are a drag ( see what I did there ? ^^) you can't really match our need for a separate max value different from the max value of the range. So rather than work with the html range input I just removed the opportunity to select a value over max & under min

app/styles/base/_slider.less
OwenCoogan: Cleaned up

tests/integration/components/o-s-s/slider-test.ts
OwenCoogan: Fixed typo

Miexil (approved):

PR #452, opened 2024-11-27
OSS::Button: Add support for `@loadingOptions` arg

phndiaye:
### What does this PR do?

This PR introduces a `loadingOptions` optional arg on the `OSS::Button` component to allow us to fine tune it. This is because we now have a case where we want to display the button's label even when loading.

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/button.ts
+++ b/addon/components/o-s-s/button.ts
@@ -74,7 +74,6 @@ interface ButtonArgs {
   skin?: string;
   size?: string;
   loading?: boolean;
-  showLabelOnLoad?: boolean;
   loadingOptions?: LoadingOptions;
   icon?: string;
   iconUrl?: string;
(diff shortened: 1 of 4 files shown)


addon/components/o-s-s/button.ts
```diff
@@ -73,6 +74,8 @@ interface ButtonArgs {
   skin?: string;
   size?: string;
   loading?: boolean;
+  showLabelOnLoad?: boolean;
```
JulienVannier66, line 77: Not used, no?

JulienVannier66 (approved):
Otherwise LGTM

addon/components/o-s-s/button.ts
phndiaye: Yep! it was my initial solution. Removed.

Miexil (approved):

OwenCoogan (approved):

PR #453, opened 2024-11-27
[vel-4374]&[vel-4376] Badge smart skin & InputContainer success and alert messages

Xegushu:
### What does this PR do?
This PR adds a smart skin to the Badge component, as well as alert and error messages to the InputContainer component

image
image

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/input-container.hbs
+++ b/addon/components/o-s-s/input-container.hbs
@@ -1,7 +1,7 @@
 <div class="fx-col">
   <div
     class="oss-input-container
-      {{if this.hasStyleKey (concat ' oss-input-container--' this.styleKey)}}
+      {{if this.hasMessage (concat ' oss-input-container--' this.MessageType)}}
       {{if (has-block 'prefix') ' has-prefix'}}
       {{if (has-block 'suffix') ' has-suffix'}}"
     ...attributes
@@ -30,15 +30,15 @@
     {{/if}}
   </div>
   {{#if @successMessage}}
-    <span class="text-color-success margin-top-px-6">
+    <span class="font-color-success-500 margin-top-px-6">
       <OSS::Icon @icon="fa-check-circle" />
       <span>{{@successMessage}}</span>
     </span>
   {{/if}}
-  {{#if @alertMessage}}
-    <span class="text-color-alert margin-top-px-6">
+  {{#if @warningMessage}}
+    <span class="font-color-warning-500 margin-top-px-6">
       <OSS::Icon @icon="fa-exclamation-circle" />
-      <span>{{@alertMessage}}</span>
+      <span>{{@warningMessage}}</span>
     </span>
   {{/if}}
   {{#if @errorMessage}}
--- a/addon/components/o-s-s/input-container.hbs
+++ b/addon/components/o-s-s/input-container.hbs
@@ -31,12 +31,12 @@
   </div>
   {{#if @errorMessage}}
     <span class="text-color-error margin-top-px-6">{{@errorMessage}}</span>
-  {{else if @feedbackMessage}}
-    <span class={{concat "margin-top-px-6 font-color-" this.feedbackMessageType "-500"}}>
-      {{#unless (eq this.feedbackMessageType "error")}}
+  {{else if this.feedbackMessage}}
+    <span class={{concat "margin-top-px-6 font-color-" this.feedbackMessage.type "-500"}}>
+      {{#unless (eq this.feedbackMessage.type "error")}}
         <OSS::Icon @icon={{this.messageIcon}} />
       {{/unless}}
-      <span>{{@feedbackMessage.value}}</span>
+      <span>{{this.feedbackMessage.value}}</span>
     </span>
   {{/if}}
 </div>
--- a/addon/components/o-s-s/input-container.hbs
+++ b/addon/components/o-s-s/input-container.hbs
@@ -1,7 +1,8 @@
 <div class="fx-col">
   <div
     class="oss-input-container
-      {{if this.hasMessage (concat ' oss-input-container--' this.MessageType)}}
+      {{if (or @errorMessage @hasError) ' oss-input-container--errored'}}
+      {{if @feedbackMessage (concat ' oss-input-container--' @feedbackMessage.type)}}
       {{if (has-block 'prefix') ' has-prefix'}}
       {{if (has-block 'suffix') ' has-suffix'}}"
     ...attributes
@@ -29,19 +30,14 @@
       <div class="suffix">{{yield to="suffix"}}</div>
     {{/if}}
   </div>
-  {{#if @successMessage}}
-    <span class="font-color-success-500 margin-top-px-6">
-      <OSS::Icon @icon="fa-check-circle" />
-      <span>{{@successMessage}}</span>
-    </span>
-  {{/if}}
-  {{#if @warningMessage}}
-    <span class="font-color-warning-500 margin-top-px-6">
-      <OSS::Icon @icon="fa-exclamation-circle" />
-      <span>{{@warningMessage}}</span>
-    </span>
-  {{/if}}
   {{#if @errorMessage}}
     <span class="text-color-error margin-top-px-6">{{@errorMessage}}</span>
+  {{else if @feedbackMessage}}
+    <span class={{concat "margin-top-px-6 font-color-" @feedbackMessage.type "-500"}}>
+      {{#unless (eq @feedbackMessage.type "error")}}
+        <OSS::Icon @icon={{this.messageIcon}} />
+      {{/unless}}
+      <span>{{@feedbackMessage.value}}</span>
+    </span>
   {{/if}}
 </div>
--- a/addon/components/o-s-s/input-container.hbs
+++ b/addon/components/o-s-s/input-container.hbs
@@ -1,7 +1,7 @@
 <div class="fx-col">
   <div
     class="oss-input-container
-      {{if this.displayErrorStyle ' oss-input-container--errored'}}
+      {{if this.hasStyleKey (concat ' oss-input-container--' this.styleKey)}}
       {{if (has-block 'prefix') ' has-prefix'}}
       {{if (has-block 'suffix') ' has-suffix'}}"
     ...attributes
@@ -29,6 +29,18 @@
       <div class="suffix">{{yield to="suffix"}}</div>
     {{/if}}
   </div>
+  {{#if @successMessage}}
+    <span class="text-color-success margin-top-px-6">
+      <OSS::Icon @icon="fa-check-circle" />
+      <span>{{@successMessage}}</span>
+    </span>
+  {{/if}}
+  {{#if @alertMessage}}
+    <span class="text-color-alert margin-top-px-6">
+      <OSS::Icon @icon="fa-exclamation-circle" />
+      <span>{{@alertMessage}}</span>
+    </span>
+  {{/if}}
   {{#if @errorMessage}}
     <span class="text-color-error margin-top-px-6">{{@errorMessage}}</span>
   {{/if}}
--- a/addon/components/o-s-s/input-container.hbs
+++ b/addon/components/o-s-s/input-container.hbs
@@ -1,8 +1,7 @@
 <div class="fx-col">
   <div
     class="oss-input-container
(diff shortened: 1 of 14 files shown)


addon/components/o-s-s/input-container.hbs
```diff
@@ -29,6 +29,18 @@
       <div class="suffix">{{yield to="suffix"}}</div>
     {{/if}}
   </div>
+  {{#if @successMessage}}
+    <span class="text-color-success margin-top-px-6">
```
phndiaye, line 33: `font-color-error-500`

addon/components/o-s-s/input-container.hbs
```diff
@@ -29,6 +29,18 @@
...
+      <OSS::Icon @icon="fa-check-circle" />
+      <span>{{@successMessage}}</span>
+    </span>
+  {{/if}}
+  {{#if @alertMessage}}
+    <span class="text-color-alert margin-top-px-6">
```
phndiaye, line 39: font-color-warning-500 instead (the text-color-X classes are legacy)

phndiaye: i guess you based it off the text-color-error which is old. just don't introduce new ones ;) 
(also don't change the error one because we most likely use it to target error messages in tests)

addon/components/o-s-s/input-container.ts
```diff
@@ -1,9 +1,13 @@
 import { action } from '@ember/object';
 import Component from '@glimmer/component';
 
+type MessageType = 'error' | 'alert' | 'success';
```
phndiaye, line 4: nit: whatabout `FeedbackMessageType` ? 🤔

app/styles/atoms/badge.less
```diff
@@ -155,3 +171,28 @@
 .upf-badge.upf-badge--extended-violet.upf-badge--plain {
   .generate-extended-palette-plain-badge(violet);
 }
+.upf-badge.upf-badge--extended-smart {
+  background: var(--color-smart-gradient-50) padding-box, var(--color-smart-gradient-500) border-box;
+  border-radius: 50em;
```
phndiaye, line 176: 50em ?? 🤔 (why that unit ??)

app/styles/atoms/input-container.less
```diff
@@ -45,10 +45,24 @@
     }
   }
 
-  &--errored {
```
phndiaye, line 48: i don't see it reappearing anywhere ? if it's a typo, are you sure it was never used anywhere? Saw that you changed at least one test below for email-input. If there are others in the rest of the platform, basically all tests targeting this will start failing after other devs update the package.

So for BC either:
- leave this as is (when `error` keep using `errored` as BEM modifier)
- keep `--errored` AND add `--error` too and we will gradually update the tests to use the new class instead
- Go hunt for all of them and fix them too (not sure you want to do that ^^)

addon/components/o-s-s/input-container.ts
```diff
@@ -1,9 +1,13 @@
...
+
 interface OSSInputContainerArgs {
   value?: string;
   disabled?: boolean;
+  successMessage?: string;
+  alertMessage?: string;
   errorMessage?: string;
```
Miexil, line 11: This works but I wonder if we would not be better off with a single param for the "message" + a param for the type (error/alert/success)

I'm thinking that if we keep with this current implementation, we will need to defined multiple tracked variables in the parent components if we ever need to handle mutliple types. Versus having a single "message" tracked var + changing the type of the messageType variable.

This is a question for everyone ^^

addon/components/o-s-s/input-container.ts
```diff
@@ -12,8 +16,15 @@ interface OSSInputContainerArgs {
 }
 
 export default class OSSInputContainer extends Component<OSSInputContainerArgs> {
-  get displayErrorStyle(): boolean {
-    return Boolean(this.args.errorMessage) || Boolean(this.args.hasError);
+  get hasStyleKey(): boolean {
+    return Boolean(this.styleKey);
+  }
+
+  get styleKey(): MessageType | null {
```
Miexil, line 23: what does `key` refer to here ? shouldn't it be `type` as implied by the `MessageType` type you created ?

app/styles/atoms/input-container.less
Miexil, line 48: Changing `errored` to `error` 100% makes sense on a grammatical level, but are you sure we will not break a lot of tests outside the `oss-components` repo ?

Xegushu: I have to admit that I didn't think about tests outside the repo
I will rollback on this

addon/components/o-s-s/input-container.ts
Miexil: If we do move on to what I am proposing, we can/should keep the existing `errorMessage` working for backwards compatibility until we do a full cleanup in the platform. (This cleanup could be a small team effort during the next cooldown for instance)

Xegushu: Good point, this would also solve the question of "what should happen when we set multiple messages at once"

addon/components/o-s-s/input-container.ts
Xegushu: I tried many configurations to make this works and stopped to the first one working, but it seems 50% works too indeed

app/styles/atoms/badge.less
Xegushu: I tried many configurations to make this works and stopped to the first one working, but it seems 50% works too indeed

addon/components/o-s-s/input-container.ts
phndiaye: +1 on this. I think it leaves the door open for more configuration on how we display messages in the future.

Also @Miexil, maybe we should have analysis cross-squad validated too?  never thought about it but it would make sense for you to see what we plan to add in x-squad repos before we actually implement it, no? 🤔

app/styles/atoms/input-container.less
Miexil: I think Phil commented on how to apply this change while preventing tests from breaking in his comment here -> https://github.com/upfluence/oss-components/pull/453#discussion_r1860830065

addon/components/o-s-s/input-container.ts
phndiaye: Indeed ahah ^^

app/styles/utilities/_borders.less
```diff
@@ -17,7 +17,7 @@
   border-color: var(--color-border-default);
 }
 
-.border-color-alert {
+.border-color-warning {
```
phndiaye, line 20: ⚠️  Also super dangerous to update this like this

Xegushu: this time I checked upfluence-wide !
this is the single occurence of this class

addon/components/o-s-s/input-container.stories.js
```diff
@@ -45,6 +45,26 @@ export default {
...
+    warningMessage: {
+      description: 'A warning message that will be displayed below the input-group.',
+      table: {
+        type: {
+          summary: 'string'
+        },
+        defaultValue: { summary: 'undefined' }
+      },
+      control: { type: 'text' }
+    },
```
phndiaye, line 67: doesn't reflect the code anymore

addon/components/o-s-s/input-container.ts
```diff
@@ -1,9 +1,16 @@
 import { action } from '@ember/object';
 import Component from '@glimmer/component';
 
+type FeedbackMessageType = 'error' | 'warning' | 'success';
+
 interface OSSInputContainerArgs {
   value?: string;
   disabled?: boolean;
+  message?: string;
```
phndiaye, line 9: what's that one? 🤔

addon/components/o-s-s/input-container.ts
```diff
@@ -1,13 +1,16 @@
...
-type MessageType = 'error' | 'alert' | 'success';
+type FeedbackMessageType = 'error' | 'warning' | 'success';
 
 interface OSSInputContainerArgs {
   value?: string;
   disabled?: boolean;
-  successMessage?: string;
-  alertMessage?: string;
+  message?: string;
+  feedbackMessage?: {
```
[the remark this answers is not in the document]
Miexil: Either you go with @phndiaye's suggestion and use only :
```
feedbackMessage: {
  type: 'error' | 'warning' | 'success';
  value: string;
}
```

or use what I suggested at the begining :
```
feedbackMessage: string;
feedbackMessageType: 'error' | 'warning' | 'success';
```

Both are fine as they are pretty much the same, we will just have to `hash` the params to use a tracked variable for the `value` with Phil's suggestion but that works too.

Yet it feels in the current code like you have a mix of both solutions 🤔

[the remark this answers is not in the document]
Miexil: Or maybe you forgot to remove `@message` ?

addon/components/o-s-s/input-container.hbs
```diff
@@ -1,7 +1,8 @@
 <div class="fx-col">
   <div
     class="oss-input-container
-      {{if this.hasStyleKey (concat ' oss-input-container--' this.styleKey)}}
+      {{if (or @errorMessage @hasError) ' oss-input-container--errored'}}
+      {{if @feedbackMessage (concat ' oss-input-container--' @feedbackMessage.type)}}
```
Miexil, line 5: Nit:
This could have probably been a bit clearer using a getter defined in the TS code instead

phndiaye (commented):
don't know at which extent it's a nitpick but i would've added some safety here to ensure the message type is supported w/ a local getter

```ts
get feedbackMessage() {
  if (this.args.feedbackMessage && ['error', 'warning', 'success'].includes(this.args.feedbackMessage.type)) {
    return this.args.feedbackMessage.type;
  }

  return undefined;
}
```

Then use `this.feedbackMessage` everywhere in the code/template so we know we cannot pass `type="meeeeh"` and it will just accept it.

Check w/ the others if it's not too nitpicky. Otherwise, lgtm beside the doc update

addon/components/o-s-s/input-container.hbs
```diff
@@ -31,5 +31,12 @@
...
   {{#if @errorMessage}}
     <span class="text-color-error margin-top-px-6">{{@errorMessage}}</span>
+  {{else if @feedbackMessage}}
+    <span class={{concat "margin-top-px-6 font-color-" this.feedbackMessageType "-500"}}>
+      {{#unless (eq this.feedbackMessageType "error")}}
+        <OSS::Icon @icon={{this.messageIcon}} />
+      {{/unless}}
+      <span>{{@feedbackMessage.value}}</span>
+    </span>
   {{/if}}
```
phndiaye, line 41: As pointed in [the comment](https://github.com/upfluence/oss-components/pull/453#pullrequestreview-2467504053), the idea was to ALWAYS use the local `this.feedbackMessage` (the snippet literally says `feedbackMessage` and not `feedbackMessageType`)

Now you have a weird situation where given `feedbackMessage` is `{type: "invalid_type", value: "my message"}`, `@feedbackMessage` will be truthy and here is what will be rendered:

```html
<span class="margin-top-px-6 font-color--500">
      <span>my message</span>
    </span>
```

- No type class (as `feedbackMessageType` returned `undefined`): the `font-color--500`
- No icon because `feedbackMessageType` returned `undefined`
- Just a message displayed in black because no valid font color class was applied...

And tests for that case would've pinpointed that

addon/components/o-s-s/input-container.ts
```diff
@@ -1,9 +1,16 @@
 import { action } from '@ember/object';
 import Component from '@glimmer/component';
 
+type FeedbackMessageType = 'error' | 'warning' | 'success';
+type FeedbackMessage = {
+  type: FeedbackMessageType;
```
Miexil, line 6: as there are not a lot of types, I think the types from `FeedbackMessageType` can directly be defined on this line.
You should also `export FeedbackMessage` so that we can use it from parent components 🙂

phndiaye (approved):
@Miexil mentioned some type export, but pre-LGTM for me ✅

Miexil (approved):

PR #473, opened 2025-02-03
Updated: OSS::PhoneNumber input with new error management parameters

Miexil:
### What does this PR do?
Adds two new parameters to OSS::PhoneNumberInput :
- @hasError : allows to mark the input as containing errors without the need of an error message
- @errorMessage: allows passing a custom error message to the input ; displays it in red along with a red border around the input

Screenshot 2025-02-03 at 14 59 34

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/phone-number-input.ts
+++ b/addon/components/o-s-s/phone-number-input.ts
@@ -1,6 +1,7 @@
 import { assert } from '@ember/debug';
 import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
+import { isBlank } from '@ember/utils';
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { countries, type CountryData } from '@upfluence/oss-components/utils/country-codes';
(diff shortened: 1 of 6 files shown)


aprentout (approved):

addon/components/o-s-s/phone-number-input.ts
```diff
@@ -10,6 +11,8 @@ interface OSSPhoneNumberInputArgs {
   prefix: string;
   number: string;
   placeholder?: string;
+  errorMessage?: string;
+  hasError?: boolean;
```
JulienVannier66, line 15: In which case can you have an error on the input without an error message?

JulienVannier66 (approved):
Otherwise LGTM

addon/components/o-s-s/phone-number-input.ts
Miexil: Sometimes during form validation we want to display only a red border ; This parameter is also defined for OSS::InputContainer 😇

PR #476, opened 2025-02-19
Empty state component

Xegushu:
### What does this PR do?
This PR adds an empty state component

image

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/empty-state.hbs
+++ b/addon/components/o-s-s/empty-state.hbs
@@ -1,4 +1,4 @@
-<div class="fx-1 fx-col fx-xalign-center fx-gap-px-18">
+<div class="fx-1 fx-col fx-xalign-center fx-gap-px-18" ...attributes>
   {{#if (has-block "image")}}
     {{yield to="image"}}
   {{else if @badgeIcon}}
--- a/addon/components/o-s-s/empty-state.hbs
+++ b/addon/components/o-s-s/empty-state.hbs
@@ -1,9 +1,18 @@
-<div>
-  <div>
-    {{@title}}
-  </div>
-  <div>
-    {{@subtitle}}
+<div class="fx-1 fx-col fx-xalign-center fx-gap-px-18">
+  {{#if (has-block "image")}}
+    <span class="">
+      {{yield to="image"}}
+    </span>
+  {{else if @image}}
+    <OSS::Badge @icon={{@image}} @size="lg" />
+  {{/if}}
+  <div class="fx-col fx-xalign-center fx-gap-px-6">
+    <div class="font-color-gray-900 font-size-{{this.titleSize}} font-weight-semibold">
+      {{@title}}
+    </div>
+    <div class="font-color-gray-500 font-size-{{this.subtitleSize}}">
+      {{@subtitle}}
+    </div>
   </div>
   {{#if (has-block "actions")}}
     <span class="">
--- a/addon/components/o-s-s/empty-state.hbs
+++ b/addon/components/o-s-s/empty-state.hbs
@@ -1,0 +1,13 @@
+<div>
+  <div>
+    {{@title}}
+  </div>
+  <div>
+    {{@subtitle}}
+  </div>
+  {{#if (has-block "actions")}}
+    <span class="">
+      {{yield to="actions"}}
+    </span>
+  {{/if}}
+</div>
--- a/addon/components/o-s-s/empty-state.hbs
+++ b/addon/components/o-s-s/empty-state.hbs
@@ -1,22 +1,18 @@
 <div class="fx-1 fx-col fx-xalign-center fx-gap-px-18">
   {{#if (has-block "image")}}
-    <span class="">
-      {{yield to="image"}}
-    </span>
-  {{else if @image}}
-    <OSS::Badge @icon={{@image}} @size="lg" />
+    {{yield to="image"}}
+  {{else if @badgeIcon}}
+    <OSS::Badge @icon={{@badgeIcon}} @size="lg" />
   {{/if}}
-  <div class="fx-col fx-xalign-center fx-gap-px-6">
+  <span class="fx-col fx-xalign-center fx-gap-px-6">
     <div class="font-color-gray-900 font-size-{{this.titleSize}} font-weight-semibold">
       {{@title}}
     </div>
-    <div class="font-color-gray-500 font-size-{{this.subtitleSize}}">
+    <div class="font-color-gray-500 font-size-{{this.size}}">
       {{@subtitle}}
     </div>
-  </div>
+  </span>
   {{#if (has-block "actions")}}
-    <span class="">
-      {{yield to="actions"}}
-    </span>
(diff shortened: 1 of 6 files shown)


addon/components/o-s-s/empty-state.hbs
```diff
@@ -0,0 +1,22 @@
+<div class="fx-1 fx-col fx-xalign-center fx-gap-px-18">
+  {{#if (has-block "image")}}
+    <span class="">
+      {{yield to="image"}}
+    </span>
+  {{else if @image}}
+    <OSS::Badge @icon={{@image}} @size="lg" />
```
phndiaye, line 7: `@badgeIcon` instead of `@image` would be better imo

addon/components/o-s-s/empty-state.hbs
```diff
@@ -0,0 +1,22 @@
+<div class="fx-1 fx-col fx-xalign-center fx-gap-px-18">
+  {{#if (has-block "image")}}
+    <span class="">
```
phndiaye, line 3: why the `span` here? as it's a named block, let's just let the caller handle the content here. same below for the actions

addon/components/o-s-s/empty-state.hbs
```diff
@@ -0,0 +1,22 @@
...
+  {{else if @image}}
+    <OSS::Badge @icon={{@image}} @size="lg" />
+  {{/if}}
+  <div class="fx-col fx-xalign-center fx-gap-px-6">
+    <div class="font-color-gray-900 font-size-{{this.titleSize}} font-weight-semibold">
+      {{@title}}
+    </div>
+    <div class="font-color-gray-500 font-size-{{this.subtitleSize}}">
+      {{@subtitle}}
+    </div>
```
phndiaye, line 15: `title` & `subtitle` are marked as optional in the TS but there is no check on their presence here, leading to empty divs.

I think we can remove the optional side in the TS and make both mandatory for now (i don't remember seeing an empty state w/ only a subtitle, or only a title)

Also, i think the `div` can be `span` instead.

addon/components/o-s-s/empty-state.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export default class OSSEmptyStateComponent extends Component<OSSEmptyStateComponentSignature> {
+  get titleSize(): string {
+    return this.args.size === 'sm' ? 'md' : 'lg';
+  }
+
+  get subtitleSize(): string {
+    return this.args.size || 'md';
+  }
+}
```
phndiaye, line 19: you could add a `size` getter that returns `this.args.size ?? 'md'` and use it locally. or even:

```
get size(): string {
  return ['sm', 'md'].includes(this.args.size) ? this.args.size : 'md';
}
```

(so we properly handle invalid sizes)

addon/components/o-s-s/empty-state.ts
```diff
@@ -0,0 +1,19 @@
...
+  subtitle: ReturnType<typeof htmlSafe>;
+  size?: 'sm' | 'md';
+}
+
+export default class OSSEmptyStateComponent extends Component<OSSEmptyStateComponentSignature> {
+  get titleSize(): string {
+    return this.size === 'sm' ? 'md' : 'lg';
+  }
+
+  get size(): string {
```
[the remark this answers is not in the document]
Xegushu: I have mixed feelings about this because it's also used for invalid values handling
Or should I add a subtitleSize getter that just pass the size getter ?

addon/components/o-s-s/empty-state.ts
```diff
@@ -0,0 +1,19 @@
+import type { htmlSafe } from '@ember/template';
+import Component from '@glimmer/component';
+
+interface OSSEmptyStateComponentSignature {
+  badgeIcon?: string;
+  title: ReturnType<typeof htmlSafe>;
+  subtitle: ReturnType<typeof htmlSafe>;
```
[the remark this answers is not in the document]
JulienVannier66: For the oss component, it can be pertinent

addon/components/o-s-s/empty-state.ts
```diff
@@ -0,0 +1,19 @@
...
+  size?: 'sm' | 'md';
+}
+
+export default class OSSEmptyStateComponent extends Component<OSSEmptyStateComponentSignature> {
+  get titleSize(): string {
+    return this.size === 'sm' ? 'md' : 'lg';
+  }
+
+  get size(): string {
+    return this.args.size && ['sm', 'md'].includes(this.args.size) ? this.args.size : 'md';
```
JulienVannier66, line 17: You can set `['sm', 'md']` in const with `ALLOWED_SIZE` for example of wording

tests/integration/components/o-s-s/empty-state-test.ts
```diff
@@ -0,0 +1,58 @@
...
+        </:actions>
+      </OSS::EmptyState>
+    `);
+
+    assert.dom('div.font-color-gray-900').hasText('No Data');
+    assert.dom('div.font-color-gray-500').hasText('Try again later');
+    assert.dom('button').hasText('Retry');
+  });
+
+  test('it applies size-based styles', async function (assert) {
```
JulienVannier66, line 53: Missing not allowed size like `random`

addon/components/o-s-s/empty-state.ts
```diff
@@ -0,0 +1,21 @@
...
+import Component from '@glimmer/component';
+
+interface OSSEmptyStateComponentSignature {
+  badgeIcon?: string;
+  title: ReturnType<typeof htmlSafe>;
+  subtitle: ReturnType<typeof htmlSafe>;
+  size?: 'sm' | 'md';
+}
+
+const ALLOWED_SIZE = ['sm', 'md'];
```
JulienVannier66, line 11:
```suggestion
const ALLOWED_SIZES = ['sm', 'md'];
```
My bad I forget the "s" at the end, don't forget the type also

addon/components/o-s-s/empty-state.ts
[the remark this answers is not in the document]
Xegushu: I added error throws and the related tests

JulienVannier66 (approved):
LGTM

addon/components/o-s-s/empty-state.stories.js
```diff
@@ -0,0 +1,95 @@
...
+        defaultValue: { summary: 'undefined' }
+      },
+      control: { type: 'select' },
+      options: ['sm', 'md']
+    }
+  },
+  parameters: {
+    docs: {
+      description: {
+        component: 'An component used when there is nothing to display on a page'
```
Miexil, line 52: Small typo here, should be `A component` instead of `an` :)

Miexil (approved):
Minor typo left in story book, pre-approving ✅

OwenCoogan (approved):

PR #480, opened 2025-03-10
Create a dedicated OSS::SearchField component

olxmpe:
### What does this PR do?
Create a dedicated OSS::SearchField component

image

### Good PR checklist
- [x] Title makes sense
- [ ] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/search-field.hbs
+++ b/addon/components/o-s-s/search-field.hbs
@@ -1,0 +1,10 @@
+<OSS::InputContainer @value={{@value}} @placeholder={{this.placeholder}} @onChange={{this.onChange}} ...attributes>
+  <:prefix>
+    <OSS::Icon @icon="fa-search" class="font-color-gray-500" />
+  </:prefix>
+  <:suffix>
+    {{#if (gt @value.length 0)}}
+      <OSS::Icon @icon="fa-times" class="font-color-gray-500" role="button" {{on "click" this.onClearSearch}} />
+    {{/if}}
+  </:suffix>
+</OSS::InputContainer>
--- a/addon/components/o-s-s/search-field.stories.js
+++ b/addon/components/o-s-s/search-field.stories.js
@@ -1,0 +1,64 @@
+import { hbs } from 'ember-cli-htmlbars';
+import { action } from '@storybook/addon-actions';
+
+export default {
+  title: 'Components/OSS::SearchField',
+  component: 'search-field',
+  argTypes: {
+    value: {
+      type: { required: true },
+      description: 'Value of the input',
+      table: {
+        type: {
+          summary: 'string'
+        },
+        defaultValue: { summary: 'undefined' }
+      },
+      control: { type: 'text' }
+    },
+    placeholder: {
+      description: 'Placeholder of the input',
+      table: {
+        type: {
+          summary: 'string'
+        },
+        defaultValue: { summary: 'undefined' }
+      },
+      control: { type: 'text' }
+    },
+    onChange: {
+      type: { required: true },
+      description: 'Method called every time the input is updated',
+      table: {
+        category: 'Actions',
+        type: {
+          summary: 'onChange(value: string): void'
+        }
+      },
+      control: { type: null }
+    }
+  },
+  parameters: {
+    docs: {
+      description: {
+        component:
+          'The OSS version of the search-field component. This component encapsulates search field input and allows user to reset the search.'
+      }
+    }
+  }
+};
+
+const defaultArgs = {
+  value: 'John',
+  placeholder: 'My placeholder',
+  onChange: action('onChange')
+};
+
+const Template = (args) => ({
+  template: hbs`
+      <OSS::SearchField @value={{this.value}} @placeholder={{this.placeholder}} @onChange={{this.onChange}} />
+  `,
+  context: args
+});
+export const BasicUsage = Template.bind({});
+BasicUsage.args = defaultArgs;
--- a/addon/components/o-s-s/search-field.ts
+++ b/addon/components/o-s-s/search-field.ts
@@ -1,0 +1,28 @@
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+import Component from '@glimmer/component';
+import type IntlService from 'ember-intl/services/intl';
+
+interface OSSSearchFieldArgs {
+  value: string;
+  placeholder?: string;
+  onChange(value: string): void;
+}
+
+export default class OSSSearchFieldComponent extends Component<OSSSearchFieldArgs> {
+  @service declare intl: IntlService;
+
+  get placeholder(): string {
+    return this.args.placeholder ?? this.intl.t('oss-components.search_field.placeholder');
+  }
+
+  @action
+  onClearSearch(): void {
+    this.args.onChange('');
+  }
+
+  @action
+  onChange(value: string): void {
+    this.args.onChange(value);
+  }
+}
--- a/tests/dummy/app/controllers/input.ts
+++ b/tests/dummy/app/controllers/input.ts
@@ -7,6 +7,7 @@ import { countries, type CountryData } from '@upfluence/oss-components/utils/cou
 export default class Input extends Controller {
   @tracked shopUrl: string = '';
   @tracked inputValue: string = '';
+  @tracked searchFieldValue: string = '';
   @tracked textAreaValue: string = '42';
   @tracked superHeroes: string[] = ['Iron Man', 'Thor', 'Loki', 'Hulk'];
   @tracked items: { name: string; label: string }[] = [
@@ -120,5 +121,10 @@ export default class Input extends Controller {
   onCurrencyOnlyChange(currency: string): void {
     console.log('onCurrencyOnlyChange', currency);
     this.currencyOnly = currency;
+  }
+
+  @action
+  onSearchFieldChange(value: string): void {
+    this.searchFieldValue = value;
   }
 }
--- a/tests/integration/components/o-s-s/search-field-test.ts
+++ b/tests/integration/components/o-s-s/search-field-test.ts
@@ -1,0 +1,58 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render, click, typeIn } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+import { setupIntl } from 'ember-intl/test-support';
+import sinon from 'sinon';
+
+module('Integration | Component | o-s-s/search-field', function (hooks) {
+  setupRenderingTest(hooks);
+  setupIntl(hooks);
+
+  hooks.beforeEach(function () {
+    this.onChange = (value: string) => {
+      this.value = value;
+    };
+  });
+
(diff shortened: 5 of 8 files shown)


addon/components/o-s-s/search-field.ts
```diff
@@ -0,0 +1,33 @@
...
+interface OSSSearchFieldArgs {
+  value: string;
+  placeholder?: string;
+  // eslint-disable-next-line no-unused-vars
+  onChange: (value: string) => void;
+}
+
+export default class OSSSearchFieldComponent extends Component<OSSSearchFieldArgs> {
+  @service declare intl: IntlService;
+  @tracked value = this.args.value;
```
phndiaye, line 16: instead of copying `this.args.value` into a local `this.value` property, it's better to always use `this.args.value` (`@value` in the template) so that the code using this component has the full ownership of the `value`.

Having a local `value` means that there is a risk of having an out-of-sync input.
For example, let's say the `onChange` function passed to this component has the following code:

```
onChange(valueFromSearchField: string) {
  this.valueFromSearchField = `${valueFromSearchField} suffix`
}
```

In that case if you search for "hello" you'll have a weird situation where:
- the parent stores `hello suffix`
- but the input still displays `hello`

addon/components/o-s-s/search-field.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  get placeholder(): string {
+    return this.args.placeholder || this.intl.t('oss-components.access-panel.search_placeholder');
```
phndiaye, line 19: nit: even if the translation contains the same text as `access-panel.search_placeholder`, i think it's better to still create a `oss-components.search_field.placeholder` translation here.

This way, if someday someone refactors or removes the `access-panel` component, they don't accidentally cause a side effect "Missing translation" here.

tests/integration/components/o-s-s/search-field-test.ts
```diff
@@ -0,0 +1,66 @@
...
+module('Integration | Component | o-s-s/search-field', function (hooks) {
+  setupRenderingTest(hooks);
+  setupIntl(hooks);
+
+  test('Search Field component is rendered with a placeholder', async function (assert) {
+    this.set('placeholder', 'This is a placeholder');
+    this.set('value', '');
+    this.set('onChange', (value: string) => {
+      this.value = value;
+    });
```
phndiaye, line 16: nit: you can directly do `this.placeholder = 'This is a placeholder'` (same for the others)

tests/integration/components/o-s-s/search-field-test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    await render(
+      hbs`<OSS::SearchField @value={{this.value}} @placeholder={{this.placeholder}} @onChange={{this.onChange}} />`
+    );
+
+    assert.strictEqual(
+      find('input')?.getAttribute('placeholder'),
+      'This is a placeholder',
+      'The placeholder is correctly set.'
+    );
```
phndiaye, line 26: Although this works fine, the `assert` module actually have a shorter way to do this:

```
assert.dom('input').hasAttribute('placeholder', 'This is a placeholder');
```

You can the see the documentation [here](https://github.com/mainmatter/qunit-dom/blob/master/API.md#hasattribute) along with some other functions that exist :)

tests/integration/components/o-s-s/search-field-test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  test('Input value is reseted when the clear button is clicked', async function (assert) {
+    this.set('value', 'Test');
+    this.set('onChange', (value: string) => {
+      this.value = value;
+    });
+
+    await render(hbs`<OSS::SearchField @value={{this.value}} @onChange={{this.onChange}} />`);
+
+    assert.ok(find('[role="button"]'), 'The clear button is shown.');
```
phndiaye, line 37: In the same vibe as in the attribute presence testing, here you can do:

```
assert.dom('[role="button"]').exists();
```

tests/integration/components/o-s-s/search-field-test.ts
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  test('onChange function is triggered on input change', async function (assert) {
+    this.set('value', '');
+    this.set('onChange', (value: string) => {
+      this.value = value;
+    });
+
+    await render(hbs`<OSS::SearchField @value={{this.value}} @onChange={{this.onChange}} />`);
+    await typeIn('input', 'Test');
```
phndiaye, line 51: for performance purpose, you can pass `{ delay: 0 }` as a 3rd parameter here.

This is because by default, the `typeIn` function adds some delay ([default: 50ms](https://github.com/emberjs/ember-test-helpers/blob/master/API.md#parameters-11)) between each character, leading to slower tests

tests/integration/components/o-s-s/search-field-test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  test('Clear button is not displayed when the input is empty', async function (assert) {
+    this.set('value', '');
+    this.set('onChange', (value: string) => {
+      this.value = value;
+    });
+
+    await render(hbs`<OSS::SearchField @value={{this.value}} @onChange={{this.onChange}} />`);
+
+    assert.notOk(find('[role="button"]'), 'The clear button is not visible when input is empty.');
```
phndiaye, line 64: With the same logic as [here](https://github.com/upfluence/oss-components/pull/480/files#diff-7e0ad3cc073d4d76f85d812af7f7d50a881f4e16b9bf1fdaa1ef592a1e3e6173R37), you have a `doesNotExist()` function too :)

addon/components/o-s-s/search-field.hbs
```diff
@@ -0,0 +1,10 @@
+<OSS::InputContainer @value={{this.value}} @placeholder={{this.placeholder}} @onChange={{this.onChange}}>
```
phndiaye, line 1: You can also add a `...attributes` here. Can be useful to pass a `data-control-name` for example :)

addon/components/o-s-s/search-field.ts
JulienVannier66, line 19:
```suggestion
    return this.args.placeholder ?? this.intl.t('oss-components.access-panel.search_placeholder');
```
If you check `undefined` or `null` value, it is better to use nullish coalescing operator

tests/dummy/app/controllers/input.ts
```diff
@@ -121,4 +122,10 @@ export default class Input extends Controller {
     console.log('onCurrencyOnlyChange', currency);
     this.currencyOnly = currency;
   }
+
+  @action
+  onSearchFieldChange(value: string): void {
+    console.log('onSearchFieldChange', value);
```
JulienVannier66, line 128: You can remove it

tests/integration/components/o-s-s/search-field-test.ts
JulienVannier66, line 16: You can defined this one in `beforeEach` block to avoid duplicating the declaration

tests/integration/components/o-s-s/search-field-test.ts
```diff
@@ -0,0 +1,66 @@
...
+    await render(hbs`<OSS::SearchField @value={{this.value}} @onChange={{this.onChange}} />`);
+
+    assert.ok(find('[role="button"]'), 'The clear button is shown.');
+
+    await click('[role="button"]');
+
+    assert.strictEqual(this.value, '', 'Input value is reseted when the clear button is clicked');
+  });
+
+  test('onChange function is triggered on input change', async function (assert) {
```
JulienVannier66, line 44: For this test, you can use `sinon` to stub the method to really test the call of the method and not the final result
```
this.onChange = sinon.stub()
```
In the assert you can do

```
assert.true(this.onChange.lastCall.calledWith('Test'))
```

tests/integration/components/o-s-s/search-field-test.ts
```diff
@@ -0,0 +1,66 @@
...
+import { render, find, click, typeIn } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+import { setupIntl } from 'ember-intl/test-support';
+
+module('Integration | Component | o-s-s/search-field', function (hooks) {
+  setupRenderingTest(hooks);
+  setupIntl(hooks);
+
+  test('Search Field component is rendered with a placeholder', async function (assert) {
+    this.set('placeholder', 'This is a placeholder');
```
JulienVannier66, line 12: Nit: You can directly set value without the `set`
```suggestion
    this.placeholder = 'This is a placeholder';
```

phndiaye (approved):

addon/components/o-s-s/search-field.stories.js
```diff
@@ -0,0 +1,62 @@
+import { hbs } from 'ember-cli-htmlbars';
+import { action } from '@storybook/addon-actions';
+
+export default {
+  title: 'Components/OSS::SearchField',
+  component: 'search-field',
+  argTypes: {
+    value: {
+      description: 'Value of the input',
```
JulienVannier66, line 9:
```suggestion
    value: {
      type: { required: true },
      description: 'Value of the input',
```
According to the ts definition this attribute is mandatory

addon/components/o-s-s/search-field.stories.js
```diff
@@ -0,0 +1,62 @@
...
+      table: {
+        type: {
+          summary: 'string'
+        },
+        defaultValue: { summary: 'undefined' }
+      },
+      control: { type: 'text' }
+    },
+    onChange: {
+      description: 'Method called every time the input is updated',
```
JulienVannier66, line 29:
```suggestion
    onChange: {
      type: { required: true },
      description: 'Method called every time the input is updated',
```
Same here

tests/integration/components/o-s-s/search-field-test.ts
```diff
@@ -0,0 +1,58 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render, click, typeIn } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+import { setupIntl } from 'ember-intl/test-support';
+import Sinon from 'sinon';
```
JulienVannier66, line 6:
```suggestion
import sinon from 'sinon';
```
We don't use the uppercase for the `sinon` definition

JulienVannier66 (approved):
Otherwise LGTM

tests/integration/components/o-s-s/search-field-test.ts
```diff
@@ -0,0 +1,58 @@
...
+    this.value = '';
+
+    await render(
+      hbs`<OSS::SearchField @value={{this.value}} @placeholder={{this.placeholder}} @onChange={{this.onChange}} />`
+    );
+
+    assert.dom('input').hasAttribute('placeholder', 'This is a placeholder');
+  });
+
+  test('Input value is reseted when the clear button is clicked', async function (assert) {
```
Miexil, line 29: Past tense of "reset" is "reset" :)

Miexil (approved):

PR #484, opened 2025-03-25
Update searchfield component to use it where needed

olxmpe:
### What does this PR do?
Update Search component Field to handle a type and use it instead of Input Container in Infininte Select and Access Panel

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [ ] Properly labeled

### DIFF
--- a/addon/components/o-s-s/infinite-select.ts
+++ b/addon/components/o-s-s/infinite-select.ts
@@ -80,9 +80,8 @@ export default class OSSInfiniteSelect extends Component<InfiniteSelectArgs> {
   }
 
   @action
-  updateSearchKeyword(event: any): void {
-    this._searchKeyword = event?.target?.value;
-
+  updateSearchKeyword(value: string): void {
+    this._searchKeyword = value;
     this.args.onSearch?.(this._searchKeyword);
   }
 
--- a/addon/components/o-s-s/search-field.hbs
+++ b/addon/components/o-s-s/search-field.hbs
@@ -12,7 +12,6 @@
       aria-labelledby={{this.placeholder}}
       {{on "input" this.onChange}}
       {{did-insert this.setupOnlyNumericListener}}
-      {{did-insert this.focusSearch}}
       {{will-destroy this.teardownOnlyNumericListener}}
     />
   </:input>
--- a/addon/components/o-s-s/search-field.hbs
+++ b/addon/components/o-s-s/search-field.hbs
@@ -1,7 +1,21 @@
-<OSS::InputContainer @value={{@value}} @placeholder={{this.placeholder}} @onChange={{this.onChange}} ...attributes>
+<OSS::InputContainer ...attributes>
   <:prefix>
     <OSS::Icon @icon="fa-search" class="font-color-gray-500" />
   </:prefix>
+  <:input>
+    <input
+      class="upf-input"
+      type={{this.type}}
+      value={{@value}}
+      autocomplete="off"
+      placeholder={{this.placeholder}}
+      aria-labelledby={{this.placeholder}}
+      {{on "input" this.onChange}}
+      {{did-insert this.setupOnlyNumericListener}}
+      {{did-insert this.focusSearch}}
+      {{will-destroy this.teardownOnlyNumericListener}}
+    />
+  </:input>
   <:suffix>
     {{#if (gt @value.length 0)}}
       <OSS::Icon @icon="fa-times" class="font-color-gray-500" role="button" {{on "click" this.onClearSearch}} />
--- a/addon/components/o-s-s/search-field.hbs
+++ b/addon/components/o-s-s/search-field.hbs
@@ -1,4 +1,8 @@
-<OSS::InputContainer ...attributes>
+<OSS::InputContainer
+  ...attributes
+  {{did-insert this.setupOnlyNumericListener}}
+  {{will-destroy this.teardownOnlyNumericListener}}
+>
   <:prefix>
     <OSS::Icon @icon="fa-search" class="font-color-gray-500" />
   </:prefix>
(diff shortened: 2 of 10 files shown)


tests/integration/components/o-s-s/access-panel-test.ts
```diff
@@ -83,7 +83,7 @@ module('Integration | Component | o-s-s/access-panel', function (hooks) {
       await renderComponent();
       await fillIn('.oss-input-container input', 'fo');
       await typeIn('.oss-input-container input', 'o', { delay: 0 });
-      assert.ok(this.onSearch.calledOnceWithExactly('foo'));
+      assert.ok(this.onSearch.calledWithExactly('foo'));
```
JulienVannier66, line 86: Why changing that in this test?

addon/components/o-s-s/search-field.ts
```diff
@@ -16,13 +18,33 @@ export default class OSSSearchFieldComponent extends Component<OSSSearchFieldArg
...
   }
 
+  get type(): string {
+    return this.args.type ?? 'text';
+  }
+
   @action
-  onClearSearch(): void {
+  onClearSearch(event: Event): void {
+    event.stopPropagation();
```
JulienVannier66, line 27: Why adding the `stopPropagation` here? What is the fixed side effect?

addon/components/o-s-s/infinite-select.ts
```diff
@@ -29,7 +29,7 @@ type InfinityItem = {
 const DEFAULT_ITEM_LABEL = 'name';
 
 export default class OSSInfiniteSelect extends Component<InfiniteSelectArgs> {
-  @tracked _searchKeyword: string = '';
+  @tracked _searchKeyword?: string;
```
JulienVannier66, line 32: Why changing that?

tests/integration/components/o-s-s/access-panel-test.ts
olxmpe: @JulienVannier66 actually changed it again, since former input value was triggered on keyup and search field component's input value is triggered on input (test was calling the function both with fo and foo, so 4 times)

JulienVannier66: `this.onSearch.lastCall.calledOnceWithExactly('foo')` can do the trick

olxmpe: Doesn't work with calledOnceWithExactly :
`Promise rejected during "it calls the onSearch arg when a keyword is typed": this.onSearch.lastCall.calledOnceWithExactly is not a function`

I changed it again for 'calledWithExactly', is it better this way ? @JulienVannier66

JulienVannier66: Yes it is good :+1:

JulienVannier66 (approved):

addon/components/o-s-s/search-field.hbs
```diff
@@ -1,7 +1,21 @@
...
+    <input
+      class="upf-input"
+      type={{this.type}}
+      value={{@value}}
+      autocomplete="off"
+      placeholder={{this.placeholder}}
+      aria-labelledby={{this.placeholder}}
+      {{on "input" this.onChange}}
+      {{did-insert this.setupOnlyNumericListener}}
+      {{did-insert this.focusSearch}}
```
Miexil, line 15: Are we confident that this will not cause some issues ? shouldn't we manually add it where needed ?

Btw, we made a modifer a while back for this `{{enable-input-autofocus}}`, described [here](https://storybook.upfluence.co/?path=/docs/helpers-modifiers-modifiers-general--page)

addon/utils/keyboard.ts
```diff
@@ -0,0 +1,38 @@
+const NUMERIC_ONLY: RegExp = /^\d$/i;
```
Miexil, line 1: I think this can be exported, it might be used from another component in the future

addon/components/o-s-s/search-field.hbs
olxmpe: Way easier with the modifier, I put it back in accessPanel template ✅ Thanks !

Miexil (approved):

addon/components/o-s-s/search-field.ts
```diff
@@ -16,13 +18,28 @@ export default class OSSSearchFieldComponent extends Component<OSSSearchFieldArg
...
+
+  setupOnlyNumericListener(element: HTMLElement): void {
+    const input: HTMLInputElement | null = element.querySelector('input');
+    input?.addEventListener('keydown', onlyNumeric);
+  }
+
+  teardownOnlyNumericListener(element: HTMLElement): void {
+    const input: HTMLInputElement | null = element.querySelector('input');
+    input?.removeEventListener('keydown', onlyNumeric);
   }
```
phndiaye, line 44: shouldn't we have a `if (this.type !== 'number') return;` at the start of these methods ? 🤔

addon/utils/keyboard.ts
phndiaye, line 1: you'll also need a `app/utils/keyboard.js` file here that has:

```
export { onlyNumeric, onlyDecimal } from '@upfluence/oss-components/utils/keyboard'`
```

this is Ember weird addon-related stuffs unfortunately :(

addon/components/o-s-s/search-field.ts
olxmpe: I had to move the function call in the template and add `@action` in the script to avoid errors, don't know if there was a better way of doing it 

&rarr; Without `@action` : this.type is undefined (tried to make it a tracked instead of a getter but same issue)
&rarr; not moving the call on the parent container in the template : `element.querySelector('input')` is undefined (but element isn't 🤔), `setupOnlyNumericListener` and `teardownOnlyNumericListener` are not triggered

phndiaye (approved):

PR #499, opened 2025-04-22
oss/toggle-switch: do not propagate event on the active part

phndiaye:
### What does this PR do?

oss/toggle-switch:
- do not propagate event on the active part
- move to .ts

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/toggle-switch.ts
+++ b/addon/components/o-s-s/toggle-switch.ts
@@ -2,8 +2,14 @@ import Component from '@glimmer/component';
 import { assert } from '@ember/debug';
 import { action } from '@ember/object';
 
-export default class OSSToggleSwitchComponent extends Component {
-  constructor(owner, args) {
+interface ToggleSwitchArgs {
+  value: boolean;
+  disabled?: boolean;
+  onChange: (value: boolean, event: Event) => void;
+}
+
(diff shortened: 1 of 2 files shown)


addon/components/o-s-s/toggle-switch.ts
```diff
@@ -18,7 +24,9 @@ export default class OSSToggleSwitchComponent extends Component {
   }
 
   @action
-  switchState(event) {
+  switchState(event: MouseEvent) {
+    event.stopPropagation();
```
phndiaye, line 28: the change initially needed ⬆️ (the rest is TS migration)

addon/components/o-s-s/toggle-switch.ts
```diff
@@ -18,7 +24,9 @@ export default class OSSToggleSwitchComponent extends Component {
   }
 
   @action
-  switchState(event) {
+  switchState(event: MouseEvent): void {
+    event.stopPropagation();
```
JulienVannier66, line 28: Maybe add an args `useCapture?: boolean`.This will add a condition to the event capture

JulienVannier66 (approved):

PR #502, opened 2025-04-25
Add support for OSS::Form component & backing FormManager service

phndiaye:
### What does this PR do?

Add support for OSS::Form component & backing FormManager service.

These two work together to properly handle forms and their validation behaviours via validator functions that are setup for fields.

In addition to them, we also have:
- a modifier allowing us to define a new field into a form along w/ its validator function
- a helper allowing us to get the current feedback set for a given field (if any)

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/form.ts
+++ b/addon/components/o-s-s/form.ts
@@ -1,0 +1,32 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+
+import type { default as FormManager, FormInstance } from '@upfluence/oss-components/services/form-manager';
+
+interface OSSFormComponentSignature {
+  onSetup(instance: FormInstance): void;
+  onSubmit(isValid: boolean): void;
+}
+
+export default class OSSFormComponent extends Component<OSSFormComponentSignature> {
+  @service declare formManager: FormManager;
+
+  @tracked declare id: string;
+  @tracked declare formInstance: FormInstance;
+
+  constructor(owner: unknown, args: OSSFormComponentSignature) {
+    super(owner, args);
+
+    this.id = this.formManager.generateId();
+    this.formInstance = this.formManager.getInstance(this.id);
+    this.args.onSetup(this.formInstance);
+  }
+
+  @action
+  onSubmit(event: SubmitEvent): void {
+    event.preventDefault();
+    this.args.onSubmit(this.formInstance.validateForm());
+  }
+}
--- a/addon/helpers/form-field-feedback.ts
+++ b/addon/helpers/form-field-feedback.ts
@@ -1,0 +1,24 @@
+import Helper from '@ember/component/helper';
+import { inject as service } from '@ember/service';
+
+import type FormManager from '../services/form-manager';
+import type { FeedbackMessage } from '@upfluence/oss-components/components/o-s-s/input-container';
+
+type Positional = string[];
+
+export interface FormFieldFeedbackSignature {
+  Args: {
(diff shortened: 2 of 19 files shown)


tests/integration/components/o-s-s/form-test.ts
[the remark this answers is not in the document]
Miexil: @phndiaye I can quickly add an extra test for this if you want.
we can probably test that the "enter" key event triggers the onSubmit method at least

addon/components/o-s-s/form.ts
```diff
@@ -0,0 +1,32 @@
...
+  onSubmit(isValid: boolean): void;
+}
+
+export default class OSSFormComponent extends Component<OSSFormComponentSignature> {
+  @service declare formManager: FormManager;
+
+  @tracked declare id: string;
+  @tracked declare formInstance: FormInstance;
+
+  constructor(owner: unknown, args: OSSFormComponentSignature) {
```
Elodie-Labbe-Upf, line 19: Is the unknown type okay here? We really can't know the type?

addon/helpers/form-field-feedback.ts
```diff
@@ -0,0 +1,24 @@
+import Helper from '@ember/component/helper';
+import { inject as service } from '@ember/service';
+
+import type FormManager from '../services/form-manager';
+import type { FeedbackMessage } from '@upfluence/oss-components/components/o-s-s/input-container';
+
+type Positional = string[];
```
Elodie-Labbe-Upf, line 7: Isn't this type a little too much? I can see that it's used twice so I guess it can be useful for that, but it doesn't feel really relevant here

Miexil (approved):
We just need to clean up the comments & do something about the Form component tests ; lgtm otherwise expected though ^^

tests/integration/components/o-s-s/form-test.ts
```diff
@@ -0,0 +1,26 @@
...
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | o-s-s/form', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function (assert) {
+    // Set any properties with this.set('myProperty', 'value');
+    // Handle any actions with this.set('myAction', function (val) { ... });
```
Elodie-Labbe-Upf, line 11: Do you keep the comments here?

tests/integration/components/o-s-s/form-test.ts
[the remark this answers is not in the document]
phndiaye: will update it, totally forgot it ahah ^^

addon/components/o-s-s/form.ts
phndiaye: Ember's teaching it like that ahah ^^ https://docs.ember-cli-typescript.com/ember/components#understanding-args.

But technically it's either an Application or an Engine instance, which are types that are not necessarily public because we shouldn't deal directly with them :)

addon/helpers/form-field-feedback.ts
phndiaye: it's only because it's repeated (and also based off other helpers ^^)

addon/components/o-s-s/form.ts
Elodie-Labbe-Upf: Okay! My Typescript soul wants to leave my body but that's okay 😂

OwenCoogan (approved):

PR #527, opened 2025-06-26
Added Smart::Immersive::Logo

OwenCoogan:
### What does this PR do?

Related to: #

Added new Immersive Logo component, which mimics some of the behavior from the logo editor in publishr

https://www.loom.com/share/c4f78fd0dd0a4936aa5b59d981856c80?sid=5f38adf9-4fcb-4d7b-a6a3-e501b117bb02

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [ ] Properly labeled


addon/components/o-s-s/smart/immersive/logo.stories.js
```diff
@@ -0,0 +1,115 @@
+import { hbs } from 'ember-cli-htmlbars';
+import { action } from '@storybook/addon-actions';
+
+export default {
+  title: 'Components/Smart::Immersive::Logo',
+  component: 'oss-smart-immersive-logo',
+  argTypes: {
+    icon: {
+      description: 'Icon name in format "name:color", e.g. "rabbit:orange"',
```
Miexil, line 9: icons & colors are defined here -> https://github.com/upfluence/upfluence-web/blob/staging/packages/publishr-admin-web/addon/models/campaign.ts#L61

I think we should store them somewhere in the project in a dedicated file (and not use the "campaign" prefix ofc)

This way we can:

use them in storybook with the proper values
Ensure the parameters passed to the component are valid (for both colors & icons)
have loops of tests over them

addon/components/o-s-s/smart/immersive/logo.ts
```diff
@@ -0,0 +1,44 @@
...
+interface OSSSmartImmersiveLogoArgs {
+  icon?: string;
+  url?: string;
+  editable: boolean;
+  loading?: boolean;
+  oversize?: boolean;
+  onClick(): void;
+}
+
+export default class OSSSmartImmersiveLogoComponent extends Component<OSSSmartImmersiveLogoArgs> {
```
Miexil, line 15: as mentioned in my previous comment, lets have assertions on the fact that the icon parameter is using the proper format (if it's the one used ofc)

app/styles/atoms/smart/logo.less
```diff
@@ -0,0 +1,236 @@
...
+
+  --logo-color-fuschia-light: #fae8ff;
+  --logo-color-fuschia-dark: #d946ef;
+
+  --logo-color-pink-light: #fce7f3;
+  --logo-color-pink-dark: #ec4899;
+
+  --logo-color-slate-light: #f1f5f9;
+  --logo-color-slate-dark: #64748b;
+}
```
Miexil, line 49: Are we sure we want to add these as variables ? 🤔 
I don't think we'll use these, as they are defined, elsewhere.

addon/components/o-s-s/smart/immersive/logo.ts
```diff
@@ -0,0 +1,44 @@
...
+  get logoColor(): string | null {
+    return this.args.icon ? `logo-icon-color_${this.args.icon.split(':')[1]}` : null;
+  }
+
+  get logoIcon(): string | null {
+    return this.args.icon?.split(':')[0] ?? null;
+  }
+
+  get dynamicImageStyle(): string | null {
+    return this.args.url ? `background-image: url(${this.args.url})` : null;
```
Miexil, line 27: looking at the design we can see a background image (picture-frame?) ; should we set it in the component as a default value when no url or icon is passed ?

app/styles/core/_smart.less
```diff
@@ -1,3 +1,4 @@
 :root {
   --border-radius-xl: 999px;
 }
+@import '../atoms/smart/logo';
```
Miexil, line 4: I think this should be with other imports instead

addon/components/o-s-s/smart/immersive/logo.stories.js
```diff
@@ -0,0 +1,130 @@
...
+    docs: {
+      description: {
+        component: 'Smart immersive logo component which supports icon or image display and editable overlay.'
+      }
+    }
+  }
+};
+
+const Template = ({ iconName, iconColor, ...rest }) => {
+  // Compose icon string
```
Miexil, line 74: forgotten prompt :trollface:

addon/components/o-s-s/smart/immersive/logo.stories.js
```diff
@@ -0,0 +1,130 @@
+import { hbs } from 'ember-cli-htmlbars';
+import { action } from '@storybook/addon-actions';
+import { LOGO_COLORS, LOGO_ICONS } from '../../../../../app/utils/logo-config';
+
+export default {
+  title: 'Components/Smart::Immersive::Logo',
+  component: 'oss-smart-immersive-logo',
+  argTypes: {
+    icon: { table: { disable: true } },
```
aprentout, line 9: Are we sure we want to disabled it, shouldn't we keep is as not editable but with a specific label to inform that this in contcatenation of 'iconName' & 'iconColor'. 
It will be less disturbing when we will go back in the doc to see how the component works 🙏

addon/components/o-s-s/smart/immersive/logo.hbs
```diff
@@ -0,0 +1,21 @@
...
+  ...attributes
+>
+  {{#if @editable}}
+    <div class="edit-overlay fx-row fx-malign-center fx-xalign-center" {{on "click" @onClick}}>
+      <OSS::Icon @icon="fa-pen" />
+    </div>
+  {{/if}}
+  {{#if (eq this.logoMode "icon")}}
+    <div class="logo-icon fx-row fx-malign-center fx-xalign-center {{this.logoColor}}">
+      <i class="fad fa-{{this.logoIcon}}"></i>
```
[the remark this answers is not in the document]
OwenCoogan: Added OSS::Icon

addon/components/o-s-s/smart/immersive/logo.stories.js
OwenCoogan: Changed & added some explicit descriptions

PR #528, opened 2025-06-27
Scrollable-panel: Handle horizontal scroll

aprentout:
### What does this PR do?

Related to: #

https://github.com/user-attachments/assets/654c5844-a5e3-47f5-867e-e6e53e7f5efe

### Good PR checklist
- [ ] Title makes sense
- [ ] Is against the correct branch
- [ ] Properly assigned
- [ ] Added/updated tests
- [ ] Properly labeled


addon/components/o-s-s/scrollable-panel.stories.js
```diff
@@ -68,5 +98,37 @@ const Template = (args) => ({
...
+
+  context: args
+});
+
 export const BasicUsage = Template.bind({});
 BasicUsage.args = defaultArgs;
+BasicUsage.argTypes = {
+  horizontal: {
+    table: {
+      disable: true
```
Miexil, line 128: I think it might be best to leave the option visible in the BasicUsage. I'm afraid we'll completely miss the option when looking into the docs and not checking the sub-usages.

Miexil (approved):

JulienVannier66 (approved):

PR #529, opened 2025-07-01
Feat: Changes to OSS::Carousel for Campaign::Card::Header needs

OwenCoogan:
### What does this PR do?

Related to: #
Capture d’écran 2025-07-01 à 14 28 34

### Good PR checklist
- [x] Title makes sense
- [] Only addresses one issue
- [x] Properly assigned
- [x] Added/updated tests
- [ ] Properly labeled

### DIFF
--- a/addon/components/o-s-s/carousel.stories.js
+++ b/addon/components/o-s-s/carousel.stories.js
@@ -67,6 +67,16 @@ export default {
           summary: 'onPageChange(): void'
         }
       }
+    },
+    showPageCounterBottom: {
+      description: 'Whether to show the page counter at the bottom of the carousel.',
+      table: {
+        type: {
+          summary: 'boolean'
+        },
+        defaultValue: { summary: 'false' }
+      },
+      control: { type: 'boolean' }
     }
   },
   parameters: {
@@ -84,14 +94,15 @@ const defaultArgs = {
   animationStyle: undefined,
   showIndicators: undefined,
   showControls: undefined,
-  autoPlay: undefined
+  autoPlay: undefined,
+  showPageCounterBottom: false
 };
 
 const Template = (args) => ({
   template: hbs`
     <OSS::Carousel @showIndicators={{this.showIndicators}} @showControls={{this.showControls}}
                    @animationStyle={{this.animationStyle}} @buttonIcon={{this.buttonIcon}}
-                   @autoPlay={{this.autoPlay}}>
+                   @autoPlay={{this.autoPlay}} @showPageCounterBottom={{this.showPageCounterBottom}}>
       <:pages>
         <div class="page">
           <OSS::Banner @icon="fas fa-image" @title="PAGE 1"/>
--- a/addon/components/o-s-s/carousel.ts
+++ b/addon/components/o-s-s/carousel.ts
@@ -9,6 +9,7 @@ interface OSSCarouselArgs {
   buttonIcon?: string;
   animationStyle?: 'shift' | 'slide';
   showIndicators?: boolean;
+  showPageCounterBottom?: boolean;
   showControls?: 'overlay' | 'outside';
   autoPlay?: number;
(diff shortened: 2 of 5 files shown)


addon/components/o-s-s/carousel.ts
```diff
@@ -9,6 +9,7 @@ interface OSSCarouselArgs {
   buttonIcon?: string;
   animationStyle?: 'shift' | 'slide';
   showIndicators?: boolean;
+  showPageCounterBottom?: boolean;
```
Miexil, line 12: the argument just above allows to hide & display the `indicators` can we keep a similar naming scheme ?
-> `indicatorsPosition: 'top' | 'bottom'` defaults to where it currently is (top I think)

Miexil: Storybook & test files will need to be updated as well

app/styles/organisms/carousel.less
```diff
@@ -6,14 +6,24 @@
...
     padding: var(--spacing-px-3);
     color: var(--color-gray-200);
-    font-size: 6px;
+    font-size: 8px;
+
+    i {
+      border: 1px solid var(--color-primary-100);
+      overflow: hidden;
+      border-radius: 999px;
+    }
```
Miexil, line 22: if we change this we will change how they already are displayed in the side-panel.
Let's not change the visuals of the indicators.

Miexil: The task mentions that there is a single usage in the soft (the sidepanel) & that we are safe to change the visuals of the `controls` not the indicators.

I understand that you checked the new visuals and are trying to match them though - but I'd rather have the product / design team update their own design based on the existing component.

app/styles/organisms/carousel.less
```diff
@@ -6,14 +6,24 @@
...
 
     &:hover,
     &--active {
-      color: var(--color-gray-500);
+      color: var(--color-primary-200);
```
Miexil, line 26: This should also be removed

app/styles/organisms/carousel.less
```diff
@@ -25,19 +35,30 @@
     display: flex;
     align-items: center;
     justify-content: center;
-    width: var(--spacing-px-24);
+    width: var(--spacing-px-36);
+    height: var(--spacing-px-36);
+    border-radius: 999px;
```
Miexil, line 40: we have a know width/height here, and it's rounded IIRC, so why not use 50% ?

OwenCoogan: Added to 50%

app/styles/organisms/carousel.less
OwenCoogan: Removed

Miexil (approved):

addon/components/o-s-s/carousel.stories.js
```diff
@@ -67,6 +67,17 @@ export default {
           summary: 'onPageChange(): void'
         }
       }
+    },
+    indicatorsPosition: {
+      description: 'Whether to show the page indicators at the bottom of the carousel.',
```
nathalieArnoux, line 72: at the bottom or the top* _(since whether expresses a choice)_

olxmpe (approved):

addon/components/o-s-s/carousel.stories.js
```diff
@@ -84,14 +95,15 @@ const defaultArgs = {
   animationStyle: undefined,
   showIndicators: undefined,
   showControls: undefined,
-  autoPlay: undefined
+  autoPlay: undefined,
+  indicatorsPosition: false
```
nathalieArnoux, line 99: Why is it a boolean here when you defined at as a type `select` on line 80?

OwenCoogan: fixed

PR #534, opened 2025-07-08
New WizardManager service along with helper classes & components

Miexil:
### Wizard

One of our upcoming projects requires a step-system in which the user can fill in some forms & move on to the next one.
Here's are sample gif & pics of the requirements:
Screen Recording 2025-03-19 at 15 45 58
Screenshot 2025-07-08 at 15 30 36

The container system allows a user to switch (vertically) from step to step. As we can see from the specifications & the designs there will be many steps to create a campaign.

From the specifications file of the project mentioned, we can gather a bunch of UX/UI Constraints which we'll need to implement:
- There are different controls for switching steps:
  - A user can click on a step label inside the menu on the left
  - They can also click on the arrow keys on the far right of the visuals
  - They should also be able to scroll within a section (if the step requires it) but also scroll out of the current step and into the next
  - A user is blocked from going to the next step if the step focused step is invalid
- The controls and actions are disabled if there is no previous step, they are also disabled on the last step of the creation flow
- Most of the steps will require a validation before allowing the user to proceed to the next one. If the validation fails, we will keep the current step focused and display error states as required
- There is a proper scrolling animation when moving from one step to another
- The focused step is vertically aligned at the center of the screen, even when it's the first step of the list
- Previous & next steps have lower opacity / gradients to fade in with the background

Along with the constraints defined in the spec file, we will also add some technical constraints:
- Steps that are neither active/focused or in previous state should not be rendered in the DOM
  - Meaning that their logic, listeners, etc. won't be running in the background
  - This is particularly important as the campaign creation flow will have 20+ steps - we don't want to overload our browsers/cpus with hundreds of event listeners everywhere !!
- We should have the ability to define each step as its own entity
  - Meaning that we can create a dedicated component for Step1, another for step2, etc. without having to play around with a huge container, or without having to pass dedicated contexts to the WizardManager service
  - Doing it this way means we'll be able to specify each step on its own & that the implementation of each step will be independant of the others (workload can be easily split during sprints)
  - This also means that each step component is independant and could very well be used in some other part of the software (campaign edition would be a good candidate for example...)
- The system managing the steps and their display should be as generic as possible
  - Meaning that we'll be able to use it elsewhere in the future and that it won't be directly linked to the campaign creation flow
  - It should accept a configuration file defining each step and the target component that the step will display (this will be clarified later on)

#### Implementation

It defines the following elements which will be described above:
- the WizardManager service & the related typings
- a BaseClass abstract class
- Wizard::Container
- Wizard::Wrapper

##### Demo

Here's a small example of how the system is running:

https://github.com/user-attachments/assets/d2cb2903-4387-4c53-a554-c6d4e7d5c415

The demo showcases the following:
- in the begining, we are showing that not all steps are fully rendered until they are either in active or preview mode
- then the scroll mechanism is demoed, note that either scrolling, clicking on the controls or on the menu allow a user to proceed to the next step (only if the step is valid...)
- some of the step hooks are logging into the console

This demo is based on the configuration file that has been added to the `wizard.ts` controller, visible in this PR.

##### WizardManager & typings

The WizardManager is the core of the step management system. Its usage will be as simple as calling the `initialize` method with a proper Configuration object.
It takes a WizardConfiguration object typed as follows:

```typescript
export type WizardConfiguration {
  options?: {
    centerStepsInContainer?: boolean;
    stepWrapperBaseClass?: string;
    containerClass?: string;
  };
  sections: {
    key: string;
    steps: {
      key: string;
      componentClass: any;
      validateStep?: () => Promise<boolean>;
    }[];
  }[];
}
```

Each section requires a key to be defined, which could, for instance, be the base of an intl key to display in a menu.
In each section, we also have an array of `Step`.

Each Step defines a key for the same reasons mentioned above. A step also has a `componentClass` attribute in which we'll pass a component instance. This component will be automatically rendered inside a `Wrapper` (defined below).
Steps can also define a `validateStep` at the configuration level. While it won't be used in our campaign creation project in the final release, it could be helpful to specify a particular behaviour at this level (see example usage above in the Demo section).

The `validateStep` is called whenever a user tries to access another step. Before accessing the next step, the `validateStep` is triggered and, depending on the output, will allow the user to move to the next step or not.
This mechanism also works in a "waterfall" fashion: if a user is on Step1 and tries to access step5, all `validateStep` validations will be triggered from step1 to step5 (excluding step5). If any validation fails, the WizardManager will automatically focus the first step that has failed.
Finally, each step has the ability to store extra attributes of any type. This could be helpful in passing specific data to the `componentClass` or just unspecified helpers required for particular usecases.

The service also has dedicated typings and getters to facilitate the implementation of components around it. They are publicly exposed in the service.

There are also some configuration options available:
- The `centerStepsInContainer` is a simple optional boolean. When truthy, it will append "empty" steps at the top & the bottom of the scrolling container in order to center on the screen the first & last real steps.
- The `stepWrapperBaseClass` allows to override the default class of the Wizard::Wrapper (defined below). It will set this base class on the container and use it as the base of the `__active`, `__previous` and `__next` classes that are applied dynamically based on the `displayState` of the step. This will be helpful for customizing the size of the steps along with the effects of the previous & next steps.
- The `containerClass` allows to override the class of the `Wizard::Container` if we want to customize the layout.

##### BaseClass abstract class

Each step we will create for the campaign-creation flow will require extending the `BaseClass` abstract class.
This will give access to the step argument and will force each step component to define methods required by the step-management system.
It also automatically registers the `validateStep` which is called by the WizardManager service everytime a user tries to go to the next step.

Extending this baseClass also gives access to some step lifecycle hooks:
- onStepSubmission: as mentioned above, used by the WizardManager service to check whether the target step is valid or not
- onFirstInsertion: triggered the first time a step becomes active and has never yet been visited
- onRevisit: triggered every time the step becomes active after already having been visited
- onLeave: The only hook which is optional and which is triggered when the target step changes from active to another state

##### Wizard::Container

This component is the one that iterates over the available steps and renders them inside Wizard::Wrapper components.
It's currently directly linked to the Wizard::Container component which iterates over the available steps and displays them properly.
It waits for the service to be properly initialized before displaying the available steps.
Each step is wrapped in a `Wizard::Wrapper` component - explained below.

While this container component currently works nicely with the `Wizard::Manager`, in the future, nothing prevents us from defining custom `container` components that match other styles (maybe we'll want a horizontal scrolling version for instance, or another that has a complete other design & animations).

##### Wizard::Wrapper

This component wraps every step in it. It is automatically called from the `Wizard::Container`, we will therefore need to use it directly.
This wrapper injects the `componentClass` component that was specified in the Configuration file. Note that each `componentClass` component is also being passed the current `@step` parameter.

The wrapper component also features a bunch of logic to smoothly handle user scrolling. Reminder: users should be able to scroll within the step, but when reaching the top or bottom, they should also be able to move to the previous or next step. Attempting to move to the next step in the wizard creation will also trigger the `validateStep` and wait for a valid response before moving to the next step.

#### Extras

As we have decided to make the wizard as generic as possible, we have also added some "example" components directly in the dummy app. There are a couple of "Step" components defined as well as a very basic menu & set of controls.

## Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/tests/unit/services/wizard-manager-test.ts
+++ b/tests/unit/services/wizard-manager-test.ts
@@ -28,7 +28,7 @@ module('Unit | Service | wizard-manager', function (hooks) {
   });
 
   module('Configuration & Initialization', function () {
-    test('Section are properly initialized', function (assert) {
+    test('Sections are properly initialized', function (assert) {
       this.service.initialize(this.config as WizardConfiguration);
 
       assert.equal(this.service.sections.length, 1);


app/styles/organisms/wizard-container.less
```diff
@@ -0,0 +1,86 @@
+.wizard-container {
+  display: flex;
+  flex-direction: column;
+  flex: 1;
+  gap: var(--spacing-px-12);
+  position: relative;
+  overflow: hidden;
+  height: 100vh;
+  padding: 30px;
```
nathalieArnoux, line 9: var(--spacing-px-30) :innocent:

app/styles/organisms/wizard-container.less
```diff
@@ -0,0 +1,86 @@
...
+  from {
+    opacity: 0;
+  }
+
+  to {
+    opacity: 1;
+  }
+}
+
+@keyframes fadeInGradientPrev {
```
nathalieArnoux, line 78: Why do we need 2 animations if they're identical? :thinking:

OwenCoogan: It might be that they are not supposed to be identical , but if one is supposed to have the opposite effect of the other we can also use `animation-direction: reverse; `

Miexil: I think I set them up as two different ones initially to handle different styles of motions.
As it's not the case anymore I'll simplify to using only one

OwenCoogan (approved):

PR #538, opened 2025-07-18
Add autocomplete arg to slider

Elodie-Labbe-Upf:
### What does this PR do?
Add autocomplete option to OSS::Slider, which is activated by default.

The Input element has a native autocomplete arg which takes only "on" or "off" values. More information in this documentation: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#off

Since the autocomplete option was needed in the OSS::Slider, the arg had to be propagated in the OSS::InputGroup used in the Slider as well as in the OSS::InputContainer used in the OSS::InputGroup.

Tests and stories have been updated to manage and showcase this new option.

Related to: #

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [ ] Properly labeled

### DIFF
--- a/addon/components/o-s-s/input-container.ts
+++ b/addon/components/o-s-s/input-container.ts
@@ -47,10 +47,7 @@ export default class OSSInputContainer extends Component<OSSInputContainerArgs> 
   }
 
   get autocomplete(): 'on' | 'off' {
-    if (!this.args.autocomplete) {
-      return 'on';
-    }
-    return AutocompleteValues.includes(this.args.autocomplete) ? this.args.autocomplete : 'on';
+    return AutocompleteValues.includes(this.args.autocomplete ?? '') ? this.args.autocomplete! : 'on';
   }
 
   @action
--- a/addon/components/o-s-s/input-container.ts
+++ b/addon/components/o-s-s/input-container.ts
@@ -19,6 +19,8 @@ interface OSSInputContainerArgs {
   onChange?(value: string): void;
 }
 
+export const AutocompleteValues = ['on', 'off'];
+
 export default class OSSInputContainer extends Component<OSSInputContainerArgs> {
   get feedbackMessage(): FeedbackMessage | undefined {
     if (this.args.feedbackMessage && ['error', 'warning', 'success'].includes(this.args.feedbackMessage.type)) {
@@ -45,7 +47,10 @@ export default class OSSInputContainer extends Component<OSSInputContainerArgs> 
   }
 
   get autocomplete(): 'on' | 'off' {
-    return this.args.autocomplete ?? 'on';
+    if (!this.args.autocomplete) {
+      return 'on';
+    }
+    return AutocompleteValues.includes(this.args.autocomplete) ? this.args.autocomplete : 'on';
   }
 
   @action
--- a/addon/components/o-s-s/input-container.ts
+++ b/addon/components/o-s-s/input-container.ts
@@ -15,6 +15,7 @@ interface OSSInputContainerArgs {
   hasError?: boolean;
   placeholder?: string;
   type?: string;
+  autocomplete?: 'on' | 'off';
   onChange?(value: string): void;
 }
 
@@ -41,6 +42,10 @@ export default class OSSInputContainer extends Component<OSSInputContainerArgs> 
 
   get type(): string {
     return this.args.type ?? 'text';
+  }
+
+  get autocomplete(): 'on' | 'off' {
+    return this.args.autocomplete ?? 'on';
   }
 
   @action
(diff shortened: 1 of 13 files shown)


JulienVannier66 (approved):
LGTM

addon/components/o-s-s/input-group.stories.js
```diff
@@ -1,6 +1,8 @@
 import { hbs } from 'ember-cli-htmlbars';
 import { action } from '@storybook/addon-actions';
 
+const AutocompleteValues = ['on', 'off'];
```
OwenCoogan, line 4: Nitty but you might want to export this to be able to use it in any other component we might want to use it , specifically those components who call the input container ( input group, input number , and others that don't come to mind at the moment ^^)

addon/components/o-s-s/input-container.ts
```diff
@@ -43,6 +44,10 @@ export default class OSSInputContainer extends Component<OSSInputContainerArgs>
     return this.args.type ?? 'text';
   }
 
+  get autocomplete(): 'on' | 'off' {
+    return this.args.autocomplete ?? 'on';
```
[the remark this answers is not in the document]
Miexil: Especially on this specific "InputContainer" component which is used everywhere

addon/components/o-s-s/slider.stories.js
```diff
@@ -1,6 +1,8 @@
 import { hbs } from 'ember-cli-htmlbars';
 import { action } from '@storybook/addon-actions';
 
+const AutocompleteValues = ['on', 'off'];
```
OwenCoogan, line 4: CF first comment, might want to leave it in input group & import it for use in slider file

addon/components/o-s-s/slider.ts
```diff
@@ -25,7 +26,8 @@ export default class SliderComponent extends Component<SliderComponentArgs> {
   sliderOptions = {
     min: this.args.min ?? 0,
     max: this.args.max ?? 100,
-    step: this.args.step ?? 1
+    step: this.args.step ?? 1,
+    autocomplete: this.args.autocomplete ?? 'on'
```
Miexil, line 30: Same question as earlier, this one might make more sense to leave "on" though. Are we sure we want to default to this value ?

Miexil (commented):
There are three components with the new parameter but only two tests files have been updated.
(And Julien has already made a comment on missing tests in a previous review )

addon/components/o-s-s/input-container.ts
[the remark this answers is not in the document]
Elodie-Labbe-Upf: The ticket specified to have it on by default. I can ask again if you think that it may be a mistake?

addon/components/o-s-s/input-group.stories.js
Elodie-Labbe-Upf: Yeah I tried importing it from another file and the storybook build fails

Elodie-Labbe-Upf: Still exporting it for .ts files, but not sure I can use it in the stories files

addon/components/o-s-s/input-container.ts
```diff
@@ -45,7 +47,10 @@ export default class OSSInputContainer extends Component<OSSInputContainerArgs>
   }
 
   get autocomplete(): 'on' | 'off' {
-    return this.args.autocomplete ?? 'on';
+    if (!this.args.autocomplete) {
+      return 'on';
+    }
```
Miexil, line 52: Is this really required ? 
Isn't it already handled with the ternary operator below ? 🤔

addon/components/o-s-s/slider.ts
```diff
@@ -27,7 +28,8 @@ export default class SliderComponent extends Component<SliderComponentArgs> {
     min: this.args.min ?? 0,
     max: this.args.max ?? 100,
     step: this.args.step ?? 1,
-    autocomplete: this.args.autocomplete ?? 'on'
+    autocomplete:
+      this.args.autocomplete && AutocompleteValues.includes(this.args.autocomplete) ? this.args.autocomplete : 'on'
```
Miexil, line 32: Same as previous comment here

maybe you are getting a ts error ? maybe defaulting back to empty string using a nullish coalescing operator in the includes value ?

tests/integration/components/o-s-s/input-group-test.ts
```diff
@@ -27,6 +27,16 @@ module('Integration | Component | o-s-s/input-group', function (hooks) {
     assert.dom('.oss-input-group-row-suffix').hasText('@domain.com');
   });
 
+  test('Passing no @autocomplete parameter works', async function (assert) {
```
Miexil, line 30: > Not passing an @autocomplete parameter defaults to "on" state ?

PR #539, opened 2025-07-22
OSS::CompletionBadge

Miexil:
### What does this PR do?
_**YABC**_ - _Yet another badge component_

The CompletionBadge component brings a new badge component with a custom progression border around it.

It extends `OSS::Badge` and thus features the same options, such as size, skin, plain, etc. (Minor updates were made to the Badge component class to have better TS typings)

The progression value can be passed as a value between 0 & 100 in the `@progress` variable; this value will then be used to apply the correct border around the component.
Not providing a value defaults to 0, and going above 100 will cap the progress at 100.

https://github.com/user-attachments/assets/9e37b3f5-b20b-4290-a37a-cf6f2c9c5935

As you can see in the video, it also supports the ability to react to the live updates of the `progress` value.
There's a code example of this in the dummy app, but basically any change to a tracked "progress" variable will be applied to the borders dynamically & with a proper animation.

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/completion-badge.ts
+++ b/addon/components/o-s-s/completion-badge.ts
@@ -2,7 +2,6 @@ import { tracked } from '@glimmer/tracking';
 import { assert } from '@ember/debug';
 import { action } from '@ember/object';
 import { htmlSafe } from '@ember/template';
-import type { SafeString } from '@ember/template/-private/handlebars';
 import { isTesting } from '@embroider/macros';
 
 import type { OSSBadgeArgs, SizeType, SkinType } from './badge';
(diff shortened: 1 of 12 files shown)


OwenCoogan (approved):

addon/components/o-s-s/completion-badge.ts
```diff
@@ -0,0 +1,114 @@
+import { tracked } from '@glimmer/tracking';
+import { assert } from '@ember/debug';
+import { action } from '@ember/object';
+import { htmlSafe } from '@ember/template';
+import type { SafeString } from '@ember/template/-private/handlebars';
```
phndiaye, line 5: this is a private type ^^ instead, use `ReturnType<typeof htmlSafe>` :)

PR #544, opened 2025-07-24
Added Smart::Tag & Smart::TagInput components

nathalieArnoux:
### What does this PR do?

Related to: #[dra-3042](https://linear.app/upfluence/issue/DRA-3042/oss-components-smarttag-and-smarttaginput)

Screenshot from 2025-07-24 17-52-24
Screenshot from 2025-07-24 17-52-35

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/smart/tag-input.hbs
+++ b/addon/components/o-s-s/smart/tag-input.hbs
@@ -1,0 +1,36 @@
+<div
+  class="tag-input-container {{if (gt this.inputValue.length 0) 'tag-input-container--active'}}"
+  {{on "click" this.onClickQueryBuilder}}
+  {{did-insert this.registerElement}}
+>
+  <div class={{this.keywordInputClasses}}>
+    {{#if @loading}}
+      <div class="animated-overlay">
+        {{or this.inputValue this.placeholder}}
+      </div>
+    {{else}}
+      <div class="fx-col">
+        <span class="hidden-span">{{if
+            (or this.inputValue this.isInputFocused)
+            this.inputValue
+            this.placeholder
+          }}</span>
+        <Input
+          class="displayed-input"
+          @value={{this.inputValue}}
+          @type="text"
+          data-control-name="tag-input"
+          {{on "keydown" this.onKeydown}}
+          {{on "focusin" this.onFocusin}}
+          {{on-click-outside this.onClickOutside}}
+        />
+      </div>
+      <div class="tag-input-empty-state">
+        {{this.placeholder}}
+      </div>
+    {{/if}}
+    <span class="tag-input-icon">
+      <OSS::Icon @style="solid" @icon="fa-arrow-turn-down-left" />
+    </span>
+  </div>
+</div>
--- a/addon/components/o-s-s/smart/tag-input.hbs
+++ b/addon/components/o-s-s/smart/tag-input.hbs
@@ -10,11 +10,9 @@
       </div>
     {{else}}
       <div class="fx-col">
-        <span class="hidden-span">{{if
-            (or this.inputValue this.isInputFocused)
-            this.inputValue
-            this.placeholder
-          }}</span>
+        <span class="hidden-span">
+          {{if (or this.inputValue this.isInputFocused) this.inputValue this.placeholder}}
+        </span>
         <Input
           class="displayed-input"
           @value={{this.inputValue}}
--- a/addon/components/o-s-s/smart/tag-input.ts
+++ b/addon/components/o-s-s/smart/tag-input.ts
@@ -42,7 +42,7 @@ export default class OSSSmartTagInput extends Component<OSSSmartTagInputArgs> {
   }
 
   get placeholder(): string {
-    return this.args.placeholder || this.intl.t('oss-components.smart.tag_input.placeholder');
+    return this.args.placeholder ?? this.intl.t('oss-components.smart.tag_input.placeholder');
   }
 
   @action
--- a/addon/components/o-s-s/smart/tag-input.ts
+++ b/addon/components/o-s-s/smart/tag-input.ts
@@ -16,6 +16,7 @@ interface OSSSmartTagInputArgs {
   onKeydown(keyword: Keyword): void;
   loading?: boolean;
   hasError?: boolean;
+  placeholder?: string;
 }
 
 export default class OSSSmartTagInput extends Component<OSSSmartTagInputArgs> {
@@ -23,7 +24,7 @@ export default class OSSSmartTagInput extends Component<OSSSmartTagInputArgs> {
 
   @tracked inputValue: string = this.args.value || '';
   @tracked isInputFocused: boolean = false;
-  @tracked declare element: HTMLElement;
+  declare element: HTMLElement;
 
   get keywordInputClasses(): string {
     const classes = ['tag-input'];
@@ -41,7 +42,7 @@ export default class OSSSmartTagInput extends Component<OSSSmartTagInputArgs> {
   }
 
   get placeholder(): string {
-    return this.intl.t('oss-components.smart.tag_input.placeholder');
+    return this.args.placeholder || this.intl.t('oss-components.smart.tag_input.placeholder');
   }
 
   @action
@@ -62,7 +63,7 @@ export default class OSSSmartTagInput extends Component<OSSSmartTagInputArgs> {
   @action
   onKeydown(event: KeyboardEvent): void {
     if (event.key === 'Enter' || event.key === 'Tab') {
-      this.saveKeyword();
+      this.formatAndNotify();
     }
   }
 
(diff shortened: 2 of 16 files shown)


addon/components/o-s-s/smart/tag-input.hbs
```diff
@@ -0,0 +1,36 @@
...
+      <div class="animated-overlay">
+        {{or this.inputValue this.placeholder}}
+      </div>
+    {{else}}
+      <div class="fx-col">
+        <span class="hidden-span">{{if
+            (or this.inputValue this.isInputFocused)
+            this.inputValue
+            this.placeholder
+          }}</span>
```
Miexil, line 17: nit: re-indent properly here

addon/components/o-s-s/smart/tag-input.ts
```diff
@@ -0,0 +1,95 @@
...
+  loading?: boolean;
+  hasError?: boolean;
+}
+
+export default class OSSSmartTagInput extends Component<OSSSmartTagInputArgs> {
+  @service declare intl: IntlService;
+
+  @tracked inputValue: string = this.args.value || '';
+  @tracked isInputFocused: boolean = false;
+  @tracked declare element: HTMLElement;
```
Miexil, line 26: nit, probably does not need to be tracked

addon/components/o-s-s/smart/tag-input.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  private saveKeyword(): void {
+    if (this.inputValue.trim().length > 0) {
+      let type: TagType = 'keyword';
+      if (this.inputValue.startsWith('@')) {
+        type = 'mention';
+      } else if (this.inputValue.startsWith('#')) {
+        type = 'hashtag';
+      }
+      let keyword = { value: this.inputValue.trim(), type };
```
Miexil, line 86: nit: const instead of let

addon/components/o-s-s/smart/tag-input.ts
```diff
@@ -0,0 +1,95 @@
...
+  @action
+  onClickOutside(_: any, event: Event): void {
+    if (!this.isInputFocused) return;
+    this.isInputFocused = false;
+    if (isBlank(this.inputValue)) return;
+    event.stopImmediatePropagation();
+    this.saveKeyword();
+  }
+
+  private saveKeyword(): void {
```
Miexil, line 78: 🤔 
nothing is being saved here right ? it's just a formatter method that then triggers the onKeydown, correct ?
Maybe formatAndNotify or something like that ?

addon/components/o-s-s/smart/tag.ts
```diff
@@ -0,0 +1,68 @@
...
+  hashtag: 'cyan',
+  mention: 'violet',
+  keyword: 'yellow'
+};
+
+const TypeIconMatch: any = {
+  hashtag: 'fa-hashtag',
+  mention: 'fa-at',
+  keyword: null
+};
```
Miexil, line 18: how about merging these two into a single definition and adding some typing ?
( also will prevent using `any` for no reason ^^ )

```typescript
export type TagType = 'keyword' | 'hashtag' | 'mention';
export type TagTypeDefinition = {
  icon?: string;
  color: string;
};

const TypeDefinition: Record<TagType, TagTypeDefinition> = {
  hashtag: {
    icon: 'fa-hashtag',
    color: 'cyan'
  },
  mention: {
    icon: 'fa-at',
    color: 'violet'
  },
  keyword: {
    color: 'yellow'
  }
};
```

addon/components/o-s-s/smart/tag.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export default class OSSSmartTag extends Component<OSSSmartTagArgs> {
+  @tracked declare element: HTMLElement;
+
+  get closable(): boolean {
+    return Boolean(this.args.onRemove);
+  }
+
+  get typeColor(): TagType {
```
Miexil, line 35: `TagType` is not returned here, `string` is. You were mislead into thinking it was correct by the `any` type you set on the TypeColorMatch initially.

addon/components/o-s-s/smart/tag.ts
```diff
@@ -0,0 +1,68 @@
...
+    return TypeColorMatch[this.args.type || 'keyword'];
```
Miexil, line 36: With the new typing suggestion I made, it would be : 
```typescript
  get typeColor(): string {
    return TypeDefinition[this.args.type ?? 'keyword'].color;
  }

  get typeIcon(): string | undefined {
    return TypeDefinition[this.args.type ?? 'keyword'].icon;
  }
```

app/styles/atoms/smart/tag-input.less
```diff
@@ -0,0 +1,103 @@
...
+
+  .tag-input {
+    width: min-content;
+    display: flex;
+    align-items: center;
+    gap: var(--spacing-px-6);
+    overflow: hidden;
+    position: relative;
+
+    * {
```
Miexil, line 20: do we really need to target all sub-elements ?
Can't we be a bit more specific ? it's used on the input class right ?

app/styles/atoms/smart/tag-input.less
```diff
@@ -0,0 +1,103 @@
...
+    130deg,
+    var(--color-gray-400) 18%,
+    var(--color-gray-300) 25%,
+    rgba(250, 198, 255, 1) 56%,
+    var(--color-primary-100) 62%,
+    var(--color-white) 66%,
+    rgba(250, 198, 255, 0.58) 68%,
+    var(--color-gray-300) 73%,
+    var(--color-gray-400) 85%
+  );
```
Miexil, line 85: @aprentout didn't you create a variable for this ? 😇

tests/integration/components/o-s-s/smart/tag-input-test.ts
```diff
@@ -0,0 +1,52 @@
...
+  });
+
+  test('it calls @onKeydown when pressing Enter', async function (assert) {
+    await render(hbs`<OSS::Smart::TagInput @value="" @onKeydown={{this.onKeydown}} />`);
+    await fillIn('[data-control-name="tag-input"]', 'foo');
+    await triggerKeyEvent('[data-control-name="tag-input"]', 'keydown', 'Enter');
+    assert.true(this.onKeydown.calledOnce, 'onKeydown is called once');
+  });
+
+  test('input is cleared after save', async function (assert) {
```
Miexil, line 35: input is cleared on validation ?

tests/integration/components/o-s-s/smart/tag-test.ts
```diff
@@ -0,0 +1,48 @@
...
+    await render(hbs`<OSS::Smart::Tag @label="foo" @onRemove={{this.onRemove}} />`);
+    assert.dom('[data-control-name="remove-tag-button"]').exists('Remove button is rendered');
+    await click('[data-control-name="remove-tag-button"]');
+    assert.ok(this.onRemove.calledOnce, 'onRemove action is called once');
+  });
+
+  test('it applies the correct size class', async function (assert) {
+    await render(hbs`<OSS::Smart::Tag @label="foo" @size="lg" />`);
+    assert.dom('.smart-tag-container').hasClass('smart-tag-container--lg', 'Has large size class');
+  });
```
Miexil, line 42: how about a module for size, with md, lg and default behaviour checked ?

tests/integration/components/o-s-s/smart/tag-test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  test('it renders with mention type', async function (assert) {
+    await render(hbs`<OSS::Smart::Tag @label="baz" @type="mention" />`);
+    assert.dom('.smart-tag').hasClass('smart-tag--color-violet', 'Has violet color for mention');
+    assert.dom('.smart-tag .fa-at').exists('Has at icon');
+    assert.dom('.smart-tag span').hasText('baz');
+  });
+
+  test('it renders a closable tag and triggers onRemove', async function (assert) {
```
Miexil, line 31: Clicking on the close button calls the `@onRemove` method

addon/components/o-s-s/smart/tag-input.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+    if (this.args.hasError) {
+      console.log('input has error');
+      classes.push('tag-input--error');
+    }
+    return classes.join(' ');
+  }
+
+  get placeholder(): string {
+    return this.intl.t('oss-components.smart.tag_input.placeholder');
```
Miexil, line 44: I think we can also add an argument for this in case we need to overload the placeholder 🙏

app/styles/atoms/smart/tag-input.less
nathalieArnoux: It's used in
- the input,
- the hidden span (so the text is the same size as the input),
- the animated overlay,
- the empty state div.
I was trying to avoid repetitions this way, but maybe it's more clear to just put the declaration in each of the involved selectors.

Miexil (approved):

PR #545, opened 2025-07-25
Remove compact notion for value inferior to 1000

JulienVannier66:
### What does this PR do?

Remove compact notion for value inferior to 1000 to avoid displaying unnecessary rounding. This allows alignment with the existing `format-numeric` helper

Before the fix with compact true:
- 859.89 -> 860
- 1256.56 -> 1.2k

After the fix with compact true:
- 859.89 -> 859.89
- 1256.56 -> 1.2k

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/helpers/format-money.ts
+++ b/addon/helpers/format-money.ts
@@ -1,6 +1,6 @@
 import Helper from '@ember/component/helper';
 
-var _getFormatter = function (currency: string, compact?: boolean) {
+let _getFormatter = function (currency: string, compact?: boolean) {
   return Intl.NumberFormat(['en-EN', 'fr-FR'], {
     style: 'currency',
     currency: currency,
@@ -10,7 +10,7 @@ var _getFormatter = function (currency: string, compact?: boolean) {
   });
 };
 
-var _formatMoney = function (
+let _formatMoney = function (
   amount: string,
   currency: string,
   format: 'raw' | 'cents' = 'raw',
@@ -22,7 +22,8 @@ var _formatMoney = function (
 };
 
 export function formatMoneyHelper(params: any[]) {
-  const [amount, currency, format = 'raw', compact = false] = params;
+  let [amount, currency, format = 'raw', compact = false] = params;
+  if ((amount < 1000 && format === 'raw') || (amount < 100000 && format === 'cents')) compact = false;
   return _formatMoney(amount, currency, format, compact);
 }
 
--- a/addon/helpers/format-money.ts
+++ b/addon/helpers/format-money.ts
@@ -1,6 +1,9 @@
 import Helper from '@ember/component/helper';
 
-let _getFormatter = function (currency: string, compact?: boolean) {
+export const PREVENT_COMPACT_NOTATION_ON_RAW_BELOW = 1000;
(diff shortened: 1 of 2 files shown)


addon/helpers/format-money.ts
```diff
@@ -22,7 +22,8 @@ var _formatMoney = function (
 };
 
 export function formatMoneyHelper(params: any[]) {
-  const [amount, currency, format = 'raw', compact = false] = params;
+  let [amount, currency, format = 'raw', compact = false] = params;
+  if ((amount < 1000 && format === 'raw') || (amount < 100000 && format === 'cents')) compact = false;
```
phndiaye, line 26: what about having a `ANTI_COMPACT_RAW_THRESHOLD = 1000` & `ANTI_COMPACT_CENTS_THRESHOLD = 100_000` consts to make this clearer?

addon/helpers/format-money.ts
```diff
@@ -1,6 +1,6 @@
 import Helper from '@ember/component/helper';
 
-var _getFormatter = function (currency: string, compact?: boolean) {
+let _getFormatter = function (currency: string, compact?: boolean) {
```
phndiaye, line 3: `let` -> `const`

addon/helpers/format-money.ts
```diff
@@ -10,7 +10,7 @@ var _getFormatter = function (currency: string, compact?: boolean) {
   });
 };
 
-var _formatMoney = function (
+let _formatMoney = function (
```
phndiaye, line 13: `let` -> `const`

addon/helpers/format-money.ts
phndiaye: `THRESHOLD` -> `CEIL` ?

addon/helpers/format-money.ts
```diff
@@ -1,6 +1,9 @@
 import Helper from '@ember/component/helper';
 
-var _getFormatter = function (currency: string, compact?: boolean) {
+const ANTI_COMPACT_RAW_THRESHOLD: number = 1000;
+const ANTI_COMPACT_CENTS_THRESHOLD: number = 100_000;
```
Miexil, line 4: Sorry but this took me a while to understand, how about
```
const PREVENT_COMPACT_NOTATION_ON_RAW_BELOW = 1000;
const PREVENT_COMPACT_NOTATION_ON_CENTS_BELOW = 100_000;
```

I know it's very long, but it leaves less room for misinterpretation IMO
😇

tests/integration/helpers/format-money-test.ts
```diff
@@ -7,25 +7,34 @@ module('Integration | Helper | format-money', function (hooks) {
...
       this.format = 'cents';
 
       await render(hbs`<div>{{format-money this.amount this.currency this.format}}</div>`);
-      assert.dom('div').hasText('$12.34');
+      assert.dom('div').hasText('$12.35');
     });
   });
 
   module('Compact formatting (K/M/B)', function () {
+    test('Before thousands value, it renders complete number when compact is passed as true', async function (assert) {
```
Miexil, line 29: nit: but you could import the value directly from the exported const in the helper ?
also `below` instead of `before`

Also I don't see the test for `CENTS`

Miexil (approved):

PR #551, opened 2025-08-13
Updated border radius variables and doc

nathalieArnoux:
### What does this PR do?

We recently updated and added new border-radius classes, and it would be best to update the documentation on storybook as well to reflect these changes.

Related to: #[dra-3515](https://linear.app/upfluence/issue/DRA-3515/update-storybook-doc-on-border-radius-classes)

Screenshot from 2025-08-13 15-00-58

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/app/styles/core/_variables.less
+++ b/app/styles/core/_variables.less
@@ -14,6 +14,8 @@
   --border-radius-sm: 8px;
   --border-radius-md: 12px;
   --border-radius-lg: 16px;
+  --border-radius-xl: 999px;
+  --border-radius-round: 50%;
 
   --topbar-height: 56px;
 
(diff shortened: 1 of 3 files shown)


app/styles/core/_variables.less
```diff
@@ -14,6 +14,8 @@
   --border-radius-sm: 8px;
   --border-radius-md: 12px;
   --border-radius-lg: 16px;
+  --border-radius-xl: 999px;
+  --border-radius-round: 50%;
```
olxmpe, line 18: Is there really a difference between xl and round ?

olxmpe (approved):

app/styles/core/_variables.less
nathalieArnoux: yes there is :) it only shows when the element's width and height are different.
Here you have 100x100px for xl and round, and 100x200px for xl and round :)
Screenshot from 2025-08-13 15-54-05

PR #558, opened 2025-08-27
Wizard container: Step properly adjust to content size

aprentout:
### What does this PR do?

Related to: #

Wizard container: 
- Adjust step size to the content
- Handle scroll behavior with adjusted content

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/wizard/container.hbs
+++ b/addon/components/wizard/container.hbs
@@ -1,9 +1,7 @@
 <div class={{or this.wizardManager.configOptions.containerClass "wizard-container"}}>
-  <div class="wizard-container__scrollable">
-    {{#if this.wizardManager.initialized}}
-      {{#each this.allSteps as |step|}}
-        <Wizard::StepWrapper @step={{step}} />
-      {{/each}}
-    {{/if}}
-  </div>
+  {{#if this.wizardManager.initialized}}
+    {{#each this.allSteps as |step|}}
+      <Wizard::StepWrapper @step={{step}} />
+    {{/each}}
+  {{/if}}
 </div>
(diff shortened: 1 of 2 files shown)


addon/components/wizard/container.hbs
```diff
@@ -1,7 +1,9 @@
 <div class={{or this.wizardManager.configOptions.containerClass "wizard-container"}}>
-  {{#if this.wizardManager.initialized}}
-    {{#each this.allSteps as |step|}}
-      <Wizard::StepWrapper @step={{step}} />
-    {{/each}}
-  {{/if}}
+  <div class="wizard-container__scrollable">
```
Miexil, line 2: as discussed yesterday, I think we might be able to remove this sub-div and use a dedicated `configOptions.containerClass` from publishr.

If it doesn't work, then we can keep this implem

Miexil (approved):

addon/components/wizard/step-wrapper.hbs
```diff
@@ -3,13 +3,12 @@
...
   data-wheel-handled={{this.wheelHandled}}
-  {{did-insert this.initStepView}}
   {{did-update this.onDisplayStateUpdate @step.displayState}}
-  {{on "scroll" this.handleScrollEvent}}
+  {{did-insert this.initStepView}}
   {{on "wheel" this.handleWheelEvent}}
   ...attributes
 >
   {{#if this.isVisible}}
-    <this.componentToDisplay @step={{@step}} />
```
aprentout, line 13: did-insert a second time here, in order to refresh the element, when component appears (size change)

Miexil (approved):

PR #560, opened 2025-09-05
Add support for expandable sidebar

phndiaye:
### What does this PR do?

Add support for expandable sidebar

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled


app/styles/organisms/sidebar.less
```diff
@@ -26,17 +31,54 @@
...
+      justify-content: flex-start;
+      padding: 0 var(--spacing-px-9);
+    }
+
+    .oss-sidebar-container__footer {
+      align-items: flex-start;
     }
   }
 
+  .oss-sidebar-container__content,
```
olxmpe, line 59: Following Julien's comment, these classes are on the same element and I think they are both used only once

olxmpe (approved):

PR #567, opened 2025-09-16
When input container is of type 'number', prevent from pasting text

olxmpe:
### What does this PR do?
When a `@type="number"` argument is passed, user is no longer able to paste anything else than numbers

Related to: #6197

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [ ] Properly labeled

### DIFF
--- a/addon/components/o-s-s/input-container.ts
+++ b/addon/components/o-s-s/input-container.ts
@@ -57,10 +57,7 @@ export default class OSSInputContainer extends Component<OSSInputContainerArgs> 
 
   @action
   _onChange(value: string): void {
-    console.log(value);
-    if (this.args.onChange) {
-      this.args.onChange(value);
-    }
+    this.args.onChange?.(value);
   }
 
   @action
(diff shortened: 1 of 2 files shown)


Elodie-DeMatteis-Upf (approved):
LGTM

phndiaye (approved):

OwenCoogan (approved):

Miexil (approved):

addon/components/o-s-s/input-container.ts
phndiaye, line 1: @olxmpe in order to fix the Firefox, case. We could also add a `beforeinput` event to prevent typing numbers altogether.

The event listener would be something like:

```javascript
handler(event) {
  if (!NUMBER_INPUT_PASSTHROUGH.includes(event.inputType) || !ALLOWED_NUMBER_INPUT.includes(event.inputType)Number.isNaN(parseInt(event.data)) || typeof parseInt(event.data) !== "number") {
     event.preventDefault();
     return;
  }
}
```

where:
- `NUMBER_INPUT_PASSTHROUGH = ['deleteContentBackward']`
- `ALLOWED_NUMBER_INPUT = ['.']
 
so people can still type decimals and delete characters

PR #572, opened 2025-09-22
WizardManager : Focus and display each step in between when changing step

Miexil:
### What does this PR do?

The current wizard-manager service implemention of `focusStep` directly sets the target step passed in parameter as the visible one. This messes up some rendering when moving to steps far enough away that their sub-components are not rendered.

The following PR makes the `focusStep` method of the wizard-manager focus each step between the current step and the target one. Which means all steps in between (and the final one ofc) are set to visible at some point.
This fixes a rendering issue with our custom "step lifecycle hooks" but also fixes when scrolling over a big amount of sections. We would only see empty steps in between, now the previews are properly rendered.

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/tests/unit/services/wizard-manager-test.ts
+++ b/tests/unit/services/wizard-manager-test.ts
@@ -130,7 +130,7 @@ module('Unit | Service | wizard-manager', function (hooks) {
 
       this.service.selectStep(step2Id, true);
       await settled();
-      assert.equal(this.service.focusedStepId, step1Id);
+      assert.equal(this.service.focusedStepId, step2Id);
       assert.true(this.validateStub.notCalled, 'validateStep was not called when bypassValidations is true');
     });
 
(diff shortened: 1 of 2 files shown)


tests/unit/services/wizard-manager-test.ts
```diff
@@ -130,7 +130,7 @@ module('Unit | Service | wizard-manager', function (hooks) {
 
       this.service.selectStep(step2Id, true);
       await settled();
-      assert.equal(this.service.focusedStepId, step1Id);
+      assert.equal(this.service.focusedStepId, step2Id);
```
Miexil, line 133: I'm not sure why these test was passing before, but the change I made here makes sense to me so 🤷

PR #620, opened 2026-01-19
Add feedback message parameter to banner component

olxmpe:
### What does this PR do?
Add a red border and error message to OSS::Banner component when `@feedbackMessage` argument is passed

Related to: #7099

Capture d’écran 2026-01-19 à 15 41 04

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [ ] Properly labeled

### DIFF
--- a/addon/components/o-s-s/banner.hbs
+++ b/addon/components/o-s-s/banner.hbs
@@ -32,7 +32,7 @@
     {{/if}}
   </div>
 
-  {{#if @feedbackMessage.value}}
-    <span class="font-color-error-500">{{@feedbackMessage.value}}</span>
+  {{#if this.feedbackMessage}}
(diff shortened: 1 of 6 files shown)


addon/components/o-s-s/banner.hbs
```diff
@@ -1,30 +1,38 @@
...
+      {{/if}}
+    </div>
+    {{#if (has-block "actions")}}
+      <div>{{yield to="actions"}}</div>
     {{/if}}
   </div>
-  {{#if (has-block "actions")}}
-    <div>{{yield to="actions"}}</div>
+
+  {{#if @feedbackMessage.value}}
```
JulienVannier66: As the value is mandatory in the `FeedbackMessage` interface, instead can you check on the `!isBlank` of the value to render the span

PR #659, opened 2026-04-22
[Groundwork] [OSS-Components] Add an animation to the `OSS::Copy` button

edouardmisset:
### What does this PR do?

Add an opt-in animation to OSS::Copy for improved feedback while preserving legacy behavior by default.

Related to: [#DRA-4998](<https://linear.app/upfluence/issue/DRA-4998/groundwork-oss-components-add-an-animation-to-the-osscopy-button>)

* OSS::Copy now supports @withAnimation to show a success checkmark animation after a successful copy.
* When @withAnimation is enabled, no success toast is shown.
* Storybook examples and tests were updated for the animated behavior.

### Developpers heads up

I can't find the related figma file in the [design system](<https://www.figma.com/design/N7SDuH9mvC9zTlsLN49N4L/2.0-UI-Kit-%F0%9F%92%A0?node-id=6118-14594&p=f&t=UrKVkfLbSoIp9G91-0>) so I couldn't update the component with the figma link. 🤷<br><br>For reviewers: to test either start the "dummy app" (`make start`) and/or start storybook (`pnpm storybook`)

### Good PR checklist

- [X] Title makes sense
- [X] Is against the correct branch
- [X] Only addresses one issue
- [X] Properly assigned
- [X] Added/updated tests
- [ ] Added/updated documentation with Figma design link. Don't forget to replace "design" by "file" in the URL. For example [https://www.figma.com/file/example](<https://www.figma.com/file/example>)
- [X] Properly labeled

### DIFF
--- a/addon/components/o-s-s/copy.stories.js
+++ b/addon/components/o-s-s/copy.stories.js
@@ -48,7 +48,7 @@ export default {
       },
       control: { type: 'text' }
     },
-    withAnimation: {
+    animated: {
       type: { name: 'boolean' },
       description: 'Enable green checkmark animation on click (3 second, no toast)',
       table: {
@@ -75,43 +75,16 @@ const defaultArgs = {
   inline: false,
   icon: undefined,
   tooltip: undefined,
-  withAnimation: false
+  animated: false
 };
 
 const BasicUsageTemplate = (args) => ({
   template: hbs`
     <div class="fx-col">
-      <OSS::Copy @value={{this.value}} @inline={{this.inline}} @icon={{this.icon}} @tooltip={{this.tooltip}} @withAnimation={{this.withAnimation}} />
+      <OSS::Copy @value={{this.value}} @inline={{this.inline}} @icon={{this.icon}} @tooltip={{this.tooltip}} @animated={{this.animated}} />
     </div>`,
   context: args
 });
 
 export const Default = BasicUsageTemplate.bind({});
 Default.args = defaultArgs;
-
-export const WithAnimation = BasicUsageTemplate.bind({});
-WithAnimation.args = {
-  ...defaultArgs,
-  withAnimation: true
-};
-WithAnimation.parameters = {
-  docs: {
-    description: {
-      story: 'Shows the green checkmark animation on click for 3 seconds. No toast notification is displayed.'
-    }
-  }
-};
-
-export const InlineWithAnimation = BasicUsageTemplate.bind({});
-InlineWithAnimation.args = {
-  ...defaultArgs,
-  inline: true,
-  withAnimation: true
-};
-InlineWithAnimation.parameters = {
-  docs: {
-    description: {
-      story: 'Inline version with the green checkmark animation on click.'
-    }
-  }
-};
--- a/addon/components/o-s-s/copy.stories.js
+++ b/addon/components/o-s-s/copy.stories.js
@@ -50,7 +50,8 @@ export default {
     },
     animated: {
       type: { name: 'boolean' },
-      description: 'Enable green checkmark animation on click (3 second, no toast)',
+      description:
+        'Enables a green checkmark animation (3 seconds) when the user clicks on the button. It also disables the success feedback toast.',
       table: {
         type: {
           summary: 'boolean'
--- a/addon/components/o-s-s/copy.stories.js
+++ b/addon/components/o-s-s/copy.stories.js
@@ -47,6 +47,17 @@ export default {
         defaultValue: { summary: 'Copy' }
       },
       control: { type: 'text' }
+    },
+    withAnimation: {
+      type: { name: 'boolean' },
+      description: 'Enable green checkmark animation on click (3 second, no toast)',
+      table: {
+        type: {
+          summary: 'boolean'
+        },
+        defaultValue: { summary: false }
+      },
+      control: { type: 'boolean' }
     }
   },
   parameters: {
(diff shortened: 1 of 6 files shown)


addon/components/o-s-s/copy.stories.js
```diff
@@ -63,16 +74,44 @@ const defaultArgs = {
...
+  inline: true,
+  withAnimation: true
+};
+InlineWithAnimation.parameters = {
+  docs: {
+    description: {
+      story: 'Inline version with the green checkmark animation on click.'
+    }
+  }
+};
```
phndiaye: considering it's just a boolean change, does it really deserve a dedicated story? (as we can just enable the boolean in Storybook and play with it)

addon/components/o-s-s/copy.ts
```diff
@@ -12,6 +12,7 @@ interface OSSCopyArgs {
   inline?: boolean;
   icon?: string;
   tooltip?: string;
+  withAnimation?: boolean;
```
phndiaye: nit: just `animate?: boolean` 🤔

addon/components/o-s-s/copy.ts
```diff
@@ -20,6 +21,8 @@ export default class OSSCopy extends Component<OSSCopyArgs> {
 
   @tracked accessibleClipboard: boolean = false;
   @tracked inline: boolean = this.args.inline ?? false;
+  @tracked withAnimation: boolean = this.args.withAnimation ?? false;
```
phndiaye: not a super fan of inline initialization based on `this.args` (vs doing it in the constructor) 

(i believe it causes issues in next Ember versions actually due to initialization happening at the same time as rendering i think)

phndiaye (approved):

OwenCoogan (approved):

addon/components/o-s-s/copy.ts
JulienVannier66: More `animated` for a boolean, no?

addon/components/o-s-s/copy.ts
```diff
@@ -64,4 +72,9 @@ export default class OSSCopy extends Component<OSSCopyArgs> {
         );
       });
   }
+
+  @action
+  resetCheckmarkOnAnimationEnd() {
```
JulienVannier66:
```suggestion
  resetCheckmarkOnAnimationEnd(): void {
```
Missing typing

tests/integration/components/o-s-s/copy-test.ts
```diff
@@ -114,5 +114,57 @@ module('Integration | Component | o-s-s/copy', function (hooks) {
...
+      hooks.beforeEach(function () {
+        sinon.stub(navigator.clipboard, 'writeText').resolves();
+      });
+
+      test('when animation is disabled, info toast is rendered', async function (assert) {
+        await render(hbs`<OSS::Copy @value="test" @withAnimation={{false}} />`);
+        await click('.upf-btn--default');
+
+        assert.true(
+          this.toastInfoStub.calledOnceWithExactly('Successfully copied to your clipboard.', 'Copied to clipboard')
```
JulienVannier66: Use the intl service directly instead of the translation value

tests/integration/components/o-s-s/copy-test.ts
```diff
@@ -114,5 +114,57 @@ module('Integration | Component | o-s-s/copy', function (hooks) {
...
+        );
+      });
+
+      test('when animation is enabled, no toast is rendered', async function (assert) {
+        await render(hbs`<OSS::Copy @value="test" @withAnimation={{true}} />`);
+        await click('.upf-btn--default');
+
+        assert.false(this.toastInfoStub.called);
```
JulienVannier66:
```suggestion
        assert.true(this.toastInfoStub.notCalled);
```
More easy to ready

addon/components/o-s-s/copy.ts
edouardmisset: I understand, I followed the pattern of the above line.
I will change them accordingly.
Thanks

app/styles/atoms/copy.less
```diff
@@ -15,3 +15,31 @@
...
+  }
+
+  90% {
+    color: var(--color-success-500);
+  }
+
+  100% {
+    opacity: 0.9;
+  }
+}
```
JulienVannier66, line 46:
```suggestion
@keyframes checkmark {
  0% {
    opacity: 0.6;
  }

  20% {
    opacity: 1;
  }

  100% {
    opacity: 0.9;
  }
}
```
The corlor animation is necessary here? Because it is always the same color `--color-success-500`

addon/components/o-s-s/copy.ts
Miexil: both would work & are shorter - sounds good to me !

JulienVannier66 (approved):

app/styles/atoms/copy.less
edouardmisset: The color changes back to the `currentColor` between 90% to 100% to avoid a flash of color (green) on the copy icon. <br>This is kind of a workaround but it works 🥲

addon/components/o-s-s/copy.stories.js
```diff
@@ -47,6 +47,17 @@ export default {
         defaultValue: { summary: 'Copy' }
       },
       control: { type: 'text' }
+    },
+    animated: {
+      type: { name: 'boolean' },
+      description: 'Enable green checkmark animation on click (3 second, no toast)',
```
Miexil: 3 seconds, disables the success toast.

app/styles/atoms/copy.less
```diff
@@ -15,3 +15,31 @@
     box-shadow: none;
   }
 }
+
+.oss-copy--animation {
+  --animation-duration: 3s;
+  will-change: color, opacity;
+  &,
```
Miexil, line 23: nit: can we add a line break here ? it will ease up readability of which item has the `animation` applied to

tests/integration/components/o-s-s/copy-test.ts
```diff
@@ -100,19 +103,77 @@ module('Integration | Component | o-s-s/copy', function (hooks) {
...
+      });
+
+      test('when animation is enabled on inline copy, checkmark is displayed', async function (assert) {
+        await render(hbs`<OSS::Copy @value="test" @inline={{true}} @animated={{true}} />`);
+
+        assert.dom('i.fa-check').doesNotExist();
+
+        await click('[data-control-name="copy-content-button"]');
+
+        assert.dom('i.fa-check').exists();
```
Miexil, line 169: can we reduce the line breaks here ?

Looks good after the render, but the others are just wasted space, no ?

addon/components/o-s-s/copy.stories.js
Miexil: Actually
`Enables a green checkmark animation (3 seconds) when the user clicks on the button. It also disables the success feedback toast.`

aprentout (approved):

Miexil (approved):

PR #660, opened 2026-04-23
Use oss infinite select option in scoped components VEL-7598

JulienVannier66:
### What does this PR do?

Use oss infinite select option in scoped components. Updated components:

- OSS::CurrencyInput
- OSS::CountrySelector
- OSS::PhoneNumberInput
- OSS::Smart::Immersive::CurrencyInput

Phone number input
Screenshot from 2026-04-23 16-01-10

Country selector
Screenshot from 2026-04-23 15-35-33
Country selector - Province
Screenshot from 2026-04-23 15-35-40

Infinite select
Screenshot from 2026-04-23 15-35-09

Currency input
Screenshot from 2026-04-23 15-34-33
Screenshot from 2026-04-23 15-34-16

Smart immersive currency input
Screenshot from 2026-04-23 15-33-38

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/country-selector.ts
+++ b/addon/components/o-s-s/country-selector.ts
@@ -129,12 +129,12 @@ export default class OSSCountrySelector extends BaseDropdown<OSSCountrySelectorA
   }
 
   @action
-  onItemSelected(value: Item, preventFocus: boolean = false): void {
-    this.closeDropdown(preventFocus);
+  onItemSelected(value: Item): void {
+    this.closeDropdown();
     this.args.onChange(value);
   }
 
   private _matchValueWithSourceList(): void {
-    if (this.selectedCountry) this.onItemSelected(this.selectedCountry, true);
+    if (this.selectedCountry) this.args.onChange(this.selectedCountry);
   }
 }
--- a/addon/components/o-s-s/country-selector.ts
+++ b/addon/components/o-s-s/country-selector.ts
@@ -47,7 +47,7 @@ export default class OSSCountrySelector extends BaseDropdown<OSSCountrySelectorA
     );
 
     if (!isEmpty(this.args.value)) {
-      scheduleOnce('afterRender', this, this._matchValueWithSourceList);
+      scheduleOnce('afterRender', this, this.matchValueWithSourceList);
     }
   }
 
@@ -129,12 +129,12 @@ export default class OSSCountrySelector extends BaseDropdown<OSSCountrySelectorA
   }
 
   @action
-  onItemSelected(value: Item): void {
-    this.closeDropdown();
+  onItemSelected(value: Item, preventFocus: boolean = false): void {
+    this.closeDropdown(preventFocus);
     this.args.onChange(value);
   }
 
-  private _matchValueWithSourceList(): void {
-    if (this.selectedCountry) this.args.onChange(this.selectedCountry);
+  private matchValueWithSourceList(): void {
+    if (this.selectedCountry) this.onItemSelected(this.selectedCountry, true);
   }
 }
--- a/addon/components/o-s-s/smart/immersive/currency-input.hbs
+++ b/addon/components/o-s-s/smart/immersive/currency-input.hbs
@@ -58,13 +58,17 @@
         {{on-click-outside this.hideCurrencySelector}}
       >
         <:option as |currency|>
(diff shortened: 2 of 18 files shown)


addon/components/o-s-s/smart/immersive/currency-input.hbs
```diff
@@ -58,13 +58,17 @@
...
-          </div>
+          <OSS::InfiniteSelect::Option
+            @title={{currency.code}}
+            @selected={{eq this.selectedCurrency currency}}
+            @onSelect={{fn this.onSelect currency}}
+          >
+            <:prefix>
+              <span class={{if (eq this.selectedCurrency currency) "font-color-primary-500"}}>
+                {{currency.symbol}}
+              </span>
```
Miexil, line 69: isn't this already handled by the `@selected` parameter of the `::Option` ?
Maybe it's not the same color ? but then that raises a question (for product or design team) - do we want to align with the rest or do we want to keep an edge case for this component ?

addon/components/o-s-s/country-selector.ts
```diff
@@ -129,12 +129,12 @@ export default class OSSCountrySelector extends BaseDropdown<OSSCountrySelectorA
   }
 
   @action
-  onItemSelected(value: Item, preventFocus: boolean = false): void {
-    this.closeDropdown(preventFocus);
+  onItemSelected(value: Item): void {
+    this.closeDropdown();
```
Miexil: are you sure we remove the `preventFocus` here ?

I see it was used before in the `_matchValueWithSourceList` but I don't recall why we needed it.

Also - could you remove the prefixing `_` in the private below ? 🙏

app/styles/base/_infinite-select.less
```diff
@@ -26,6 +30,7 @@
 
 .upf-infinite-select__container {
   position: relative;
+  overflow: auto;
```
Miexil, line 33: No unforeseen impact of this elsewhere ?
Did you already check by linking to upf-web ?

addon/components/o-s-s/smart/immersive/currency-input.hbs
aprentout: I'm not sure there is current usage of the this `prefix` name block, as Max suggest maybe we need to quick sync with design to ensure this behavior could be the default one

app/styles/base/_infinite-select.less
JulienVannier66: Yes it is ok in upfluence-web

Account settings
Screenshot from 2026-04-24 14-38-32

Side panel - extra payment
Screenshot from 2026-04-24 14-37-41

Side panel - influence information
Screenshot from 2026-04-24 14-37-22

2FA number update

[Screencast from 2026-04-24 14-39-33.webm](https://github.com/user-attachments/assets/d6643c91-e669-4903-a80b-77598ef0c783)

addon/components/o-s-s/smart/immersive/currency-input.hbs
JulienVannier66: I ask to a design for currency to be sure, waiting design response

JulienVannier66: Here the design for currency input, the symbol need to have the color too. The problem is we can't set has icon (it is not a fontawesome icon). I don't have an other idea

image

addon/components/o-s-s/country-selector.ts
JulienVannier66: Roolback it, I didn't notice any difference, but it seems like a workaround for a specific case

addon/components/o-s-s/smart/immersive/currency-input.hbs
aprentout: Based on the product response it could be great to have this value defined in [infinite-select-option.less](https://github.com/upfluence/oss-components/blob/ef6e4061a0e7808a1feb832b005116c7cc865dbc/app/styles/molecules/infinite-select-option.less#L70-L72) file directly

aprentout (approved):

Miexil (approved):

olxmpe (approved):

PR #662, opened 2026-04-27
[Groundwork] [OSS-Components] Create an `OSS::FeatureCard` component & `OSS::FeatureCardsContainer`

edouardmisset:
### What does this PR do?

Implements DRA-4999 by introducing:

- OSS::FeatureCard component
- OSS::FeatureCardsContainer component that composes 2 or 3 cards with automatic colors, shadows, angles, overlap, and z-index

Related to: [#4999](https://linear.app/upfluence/issue/DRA-4999/groundwork-oss-components-create-an-ossfeaturecard-component-and)

Frontend:

- New single card component with title, description, image, color variant, and shadow variant.
- New cards container that accepts two or three cards and applies layout rules
- Added visual examples in Storybook:
  - Components/OSS::FeatureCard
  - Components/OSS::FeatureCardsContainer
- Added visual examples in the dummy app visual page.

image

Figma design:
<https://www.figma.com/design/xd32CWO1lHvIfbJ7ZTZJxq/%E2%9A%99%EF%B8%8F-Provide-OAuth-Client-Credentials-Authentication-for-API-Customers?node-id=6304-44946&t=bbpU8LxBAMqvodAV-4>

### Developers heads up

N/A

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/feature-card.hbs
+++ b/addon/components/o-s-s/feature-card.hbs
@@ -1,7 +1,7 @@
 <div class={{this.computedClasses}} ...attributes>
   <div class="oss-feature-card__content">
     <h3 class="oss-feature-card__title">{{@title}}</h3>
-    <p class="oss-feature-card__description">{{@description}}</p>
+    <p class="oss-feature-card__description text-ellipsis">{{@description}}</p>
   </div>
 
   <img class="oss-feature-card__illustration" src={{@image.src}} alt={{this.imageAlt}} />
--- a/addon/components/o-s-s/feature-card.hbs
+++ b/addon/components/o-s-s/feature-card.hbs
@@ -1,12 +1,8 @@
-<div
-  class="oss-feature-card oss-feature-card--color-{{this.colorVariant}} oss-feature-card--shadow-{{this.shadowVariant}}"
-  ...attributes
->
+<div class={{this.computedClasses}} ...attributes>
   <div class="oss-feature-card__content">
     <h3 class="oss-feature-card__title">{{@title}}</h3>
     <p class="oss-feature-card__description">{{@description}}</p>
   </div>
 
   <img class="oss-feature-card__illustration" src={{@image.src}} alt={{this.imageAlt}} />
-
 </div>
--- a/addon/components/o-s-s/feature-card.hbs
+++ b/addon/components/o-s-s/feature-card.hbs
@@ -1,0 +1,12 @@
+<div
+  class="oss-feature-card oss-feature-card--color-{{this.colorVariant}} oss-feature-card--shadow-{{this.shadowVariant}}"
+  ...attributes
+>
+  <div class="oss-feature-card__content">
+    <h3 class="oss-feature-card__title">{{@title}}</h3>
+    <p class="oss-feature-card__description">{{@description}}</p>
+  </div>
+
+  <img class="oss-feature-card__illustration" src={{@image.src}} alt={{this.imageAlt}} />
+
+</div>
--- a/addon/components/o-s-s/feature-card.hbs
+++ b/addon/components/o-s-s/feature-card.hbs
@@ -1,6 +1,6 @@
 <div class={{this.computedClasses}} ...attributes>
   <div class="oss-feature-card__content">
-    <h3 class="oss-feature-card__title">{{@title}}</h3>
+    <span class="font-size-md font-weight-semibold">{{@title}}</span>
     <p class="oss-feature-card__description text-ellipsis">{{@description}}</p>
   </div>
 
--- a/addon/components/o-s-s/feature-card.stories.js
+++ b/addon/components/o-s-s/feature-card.stories.js
@@ -75,13 +75,13 @@ const defaultArgs = {
 
 const Template = (args) => ({
   template: hbs`
-      <OSS::FeatureCard
-        @title={{this.title}}
-        @description={{this.description}}
-        @image={{this.image}}
-        @colorVariant={{this.colorVariant}}
-        @shadowVariant={{this.shadowVariant}}
-      />
+    <OSS::FeatureCard
+      @title={{this.title}}
+      @description={{this.description}}
+      @image={{this.image}}
+      @colorVariant={{this.colorVariant}}
+      @shadowVariant={{this.shadowVariant}}
+    />
   `,
   context: args
 });
--- a/addon/components/o-s-s/feature-card.stories.js
+++ b/addon/components/o-s-s/feature-card.stories.js
@@ -1,0 +1,90 @@
+import { hbs } from 'ember-cli-htmlbars';
+
+const ColorVariants = ['blue', 'violet', 'yellow'];
+const ShadowVariants = ['sm', 'lg'];
+
+export default {
+  title: 'Components/OSS::FeatureCard',
+  component: 'feature-card',
+  argTypes: {
+    title: {
+      description: 'Card title',
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: 'undefined' }
+      },
+      control: { type: 'text' },
+      type: { required: true }
+    },
+    description: {
+      description: 'Card description',
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: 'undefined' }
+      },
+      control: { type: 'text' },
+      type: { required: true }
+    },
+    image: {
+      description: 'Image object including src and optional alt',
+      table: {
+        type: { summary: '{ src: string; alt?: string }' },
+        defaultValue: { summary: 'undefined' }
+      },
+      control: { type: 'object' },
+      type: { required: true }
+    },
+    colorVariant: {
+      description: 'Card background color variant',
+      table: {
+        type: { summary: ColorVariants.join('|') },
+        defaultValue: { summary: 'violet' }
+      },
+      options: ColorVariants,
+      control: { type: 'select' }
+    },
+    shadowVariant: {
+      description: 'Card shadow variant',
+      table: {
+        type: { summary: ShadowVariants.join('|') },
+        defaultValue: { summary: 'sm' }
+      },
+      options: ShadowVariants,
+      control: { type: 'select' }
+    }
+  },
+  parameters: {
+    docs: {
+      description: {
+        component: 'Displays a card with title, description and illustration. It includes color and shadow variants.'
+      }
+    }
+  }
+};
+
+const defaultArgs = {
+  title: 'Audience & content insights',
+  description: 'Pull demographics and media performance into your BI to target smarter and report faster.',
+  image: {
+    src: '/@upfluence/oss-components/assets/images/no-image.svg',
+    alt: 'No image illustration'
+  },
+  colorVariant: 'violet',
+  shadowVariant: 'lg'
+};
+
+const Template = (args) => ({
+  template: hbs`
+      <OSS::FeatureCard
+        @title={{this.title}}
+        @description={{this.description}}
+        @image={{this.image}}
+        @colorVariant={{this.colorVariant}}
+        @shadowVariant={{this.shadowVariant}}
+      />
+  `,
+  context: args
+});
+
+export const Default = Template.bind({});
+Default.args = defaultArgs;
--- a/addon/components/o-s-s/feature-card.ts
+++ b/addon/components/o-s-s/feature-card.ts
@@ -60,6 +60,14 @@ export default class OSSFeatureCard extends Component<OSSFeatureCardArgs> {
     return this.args.shadowVariant ?? DEFAULT_SHADOW_VARIANT;
   }
 
+  get computedClasses(): string {
+    return [
+      'oss-feature-card',
+      `oss-feature-card--color-${this.colorVariant}`,
+      `oss-feature-card--shadow-${this.shadowVariant}`
+    ].join(' ');
+  }
+
   get imageAlt(): string {
     return this.args.image.alt ?? '';
   }
--- a/addon/components/o-s-s/feature-card.ts
+++ b/addon/components/o-s-s/feature-card.ts
@@ -1,5 +1,6 @@
 import Component from '@glimmer/component';
 import { assert } from '@ember/debug';
+import { isSafeString } from '@upfluence/oss-components/utils';
 
 export const COLOR_VARIANTS = ['blue', 'violet', 'yellow'] as const;
 export type OSSFeatureCardColorVariant = (typeof COLOR_VARIANTS)[number];
@@ -28,11 +29,11 @@ export default class OSSFeatureCard extends Component<OSSFeatureCardArgs> {
 
     assert(
       '[OSS::FeatureCard] The @title parameter is mandatory',
-      typeof args.title === 'string' && args.title.trim().length > 0
+      typeof (args.title === 'string' || isSafeString(args.title)) && args.title.trim().length > 0
     );
     assert(
       '[OSS::FeatureCard] The @description parameter is mandatory',
-      typeof args.description === 'string' && args.description.trim().length > 0
+      typeof (args.description === 'string' || isSafeString(args.description)) && args.description.trim().length > 0
     );
     assert(
       '[OSS::FeatureCard] The @image parameter is mandatory and must contain a src key',
--- a/addon/components/o-s-s/feature-card.ts
+++ b/addon/components/o-s-s/feature-card.ts
@@ -1,0 +1,66 @@
+import Component from '@glimmer/component';
+import { assert } from '@ember/debug';
+
+export const COLOR_VARIANTS = ['blue', 'violet', 'yellow'] as const;
+export type OSSFeatureCardColorVariant = (typeof COLOR_VARIANTS)[number];
+const DEFAULT_COLOR_VARIANT = 'violet' as const satisfies OSSFeatureCardColorVariant;
+
+export const SHADOW_VARIANTS = ['sm', 'lg'] as const;
+export type OSSFeatureCardShadowVariant = (typeof SHADOW_VARIANTS)[number];
+const DEFAULT_SHADOW_VARIANT = 'sm' as const satisfies OSSFeatureCardShadowVariant;
+
+export type OSSFeatureCardImage = {
+  src: string;
+  alt?: string;
+};
+
+export type OSSFeatureCardArgs = {
+  title: string;
+  description: string;
+  image: OSSFeatureCardImage;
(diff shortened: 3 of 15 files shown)


addon/components/o-s-s/feature-card.ts
```diff
@@ -0,0 +1,66 @@
...
+    super(owner, args);
+
+    assert(
+      '[OSS::FeatureCard] The @title parameter is mandatory',
+      typeof args.title === 'string' && args.title.trim().length > 0
+    );
+    assert(
+      '[OSS::FeatureCard] The @description parameter is mandatory',
+      typeof args.description === 'string' && args.description.trim().length > 0
+    );
```
phndiaye, line 39: if you pass a translation via `Intl#t` with `htmlSafe = true`, `typeof args.description` isn't `string` anymore but `SafeString` and this will blow up 💥

there is an util to add that check too to the conditions which you can find an example of in https://github.com/upfluence/oss-components/blob/8e387f9a3a3f315373aec36c66c53196e02629a2/addon/components/o-s-s/onboarding-state.ts#L16-L19

app/styles/molecules/feature-card.less
```diff
@@ -0,0 +1,64 @@
...
+  display: flex;
+  flex-direction: column;
+  gap: var(--spacing-px-24);
+  align-items: flex-start;
+  justify-content: center;
+  max-inline-size: 342px;
+  padding: var(--spacing-px-30) var(--spacing-px-60);
+  border: 1px solid;
+  border-radius: var(--border-radius-md);
+  font-family: var(--font-family-base);
```
phndiaye: This one isn't in the codebase 🤔

addon/components/o-s-s/feature-card.hbs
```diff
@@ -0,0 +1,12 @@
...
+  class="oss-feature-card oss-feature-card--color-{{this.colorVariant}} oss-feature-card--shadow-{{this.shadowVariant}}"
+  ...attributes
+>
+  <div class="oss-feature-card__content">
+    <h3 class="oss-feature-card__title">{{@title}}</h3>
+    <p class="oss-feature-card__description">{{@description}}</p>
+  </div>
+
+  <img class="oss-feature-card__illustration" src={{@image.src}} alt={{this.imageAlt}} />
+
```
Miexil: extra line-break

addon/components/o-s-s/feature-card.hbs
```diff
@@ -0,0 +1,12 @@
+<div
+  class="oss-feature-card oss-feature-card--color-{{this.colorVariant}} oss-feature-card--shadow-{{this.shadowVariant}}"
+  ...attributes
+>
+  <div class="oss-feature-card__content">
+    <h3 class="oss-feature-card__title">{{@title}}</h3>
```
Miexil: from the figma file, title font-size looks to be 14px and description 12px

I checked and it seems like our h3 tag overload puts font-size at `font-size: 24/@rem;` which is not 14px ?

addon/components/o-s-s/feature-card.stories.js
```diff
@@ -0,0 +1,90 @@
...
+    src: '/@upfluence/oss-components/assets/images/no-image.svg',
+    alt: 'No image illustration'
+  },
+  colorVariant: 'violet',
+  shadowVariant: 'lg'
+};
+
+const Template = (args) => ({
+  template: hbs`
+      <OSS::FeatureCard
```
Miexil: nit, looks like there's an extra indentation tab

app/styles/molecules/feature-card.less
```diff
@@ -0,0 +1,64 @@
+.oss-feature-card {
+  display: flex;
+  flex-direction: column;
+  gap: var(--spacing-px-24);
+  align-items: flex-start;
+  justify-content: center;
+  max-inline-size: 342px;
+  padding: var(--spacing-px-30) var(--spacing-px-60);
+  border: 1px solid;
```
JulienVannier66, line 9: No corlor border?

app/styles/molecules/feature-card.less
```diff
@@ -0,0 +1,64 @@
...
+  border-radius: var(--border-radius-md);
+  font-family: var(--font-family-base);
+  line-height: 1.6;
```
JulienVannier66: Defined like this in figma?

addon/components/o-s-s/feature-card.hbs
```diff
@@ -0,0 +1,12 @@
+<div
+  class="oss-feature-card oss-feature-card--color-{{this.colorVariant}} oss-feature-card--shadow-{{this.shadowVariant}}"
```
Miexil: nit: might look better to have these in a `computedClasses` getter in the component class

app/styles/molecules/feature-card.less
```diff
@@ -0,0 +1,64 @@
...
+  &__content {
+    display: flex;
+    flex-direction: column;
+    gap: var(--spacing-px-3);
+  }
+
+  &__title {
+    font-size: var(--font-size-md);
+    font-weight: var(--font-weight-semibold);
+    color: var(--color-gray-900);
```
JulienVannier66: Do you need to precise this value? Not at this color by default?

tests/dummy/app/controllers/visual.ts
```diff
@@ -112,6 +113,44 @@ export default class Visual extends Controller {
   @tracked feedbackMessageWarning: FeedbackMessage = { type: 'warning', value: '' };
   @tracked feedbackMessageSuccess: FeedbackMessage = { type: 'success', value: '' };
 
+  @tracked featureCardImage: OSSFeatureCardImage = {
```
JulienVannier66: Nit: Don't need to by tracked if you don't update it, same for the varaible below

addon/components/o-s-s/feature-cards-container.hbs
```diff
@@ -0,0 +1,16 @@
+<div class="oss-feature-cards-container" ...attributes>
+  {{#each this.cardsWithLayout as |card|}}
+    <div
+      class="oss-feature-cards-container__item oss-feature-cards-container__item{{if card.isCenter '--center'}}"
```
Miexil: looks like there would be a duplicate `oss-feature-cards-container__item` if the condition is not matched.
How about wrapping the class in the condition directly ?
```
{{if card.isCenter 'oss-feature-cards-container__item--center'}}
```

Could also look better in a dedicated getter in the component class though

addon/components/o-s-s/feature-card.ts
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+
+  get colorVariant(): OSSFeatureCardColorVariant {
+    return this.args.colorVariant ?? DEFAULT_COLOR_VARIANT;
+  }
+
+  get shadowVariant(): OSSFeatureCardShadowVariant {
+    return this.args.shadowVariant ?? DEFAULT_SHADOW_VARIANT;
+  }
```
JulienVannier66, line 64: What happened if the set a wrong color or shadow value?

addon/components/o-s-s/feature-cards-container.hbs
JulienVannier66:
```suggestion
      class="oss-feature-cards-container__item {{concat 'oss-feature-cards-container__item' (if card.isCenter '--center')}}"
```
Nit: More easy to read

addon/components/o-s-s/feature-cards-container.ts
```diff
@@ -0,0 +1,92 @@
...
+  isCenter: boolean;
+  style: string;
+};
+
+const TWO_CARDS_OFFSET_X = '45%';
+const THREE_CARD_OFFSET_X = '80%';
+const ROTATION_ANGLE = 11.25;
+
+const CARDS_LAYOUT = {
+  2: [
```
JulienVannier66: Why this structure with `2: [...], 3: [...]`, the explicite number in necessary?

addon/components/o-s-s/feature-cards-container.ts
```diff
@@ -0,0 +1,92 @@
...
+    { colorVariant: 'violet', shadowVariant: 'lg', rotation: 0, offsetX: '0', isCenter: true },
+    {
+      colorVariant: 'yellow',
+      shadowVariant: 'sm',
+      rotation: ROTATION_ANGLE,
+      offsetX: THREE_CARD_OFFSET_X,
+      isCenter: false
+    }
+  ]
+} as const satisfies Record<
```
Miexil: I don't think this component should be responsible of setting the card colors.

I feel like the `cards` array parameter should just take the exact same parameter set as the `FeatureCard` component.

Only Rotation should be handled by this container component and eventually have defaults for the shadows

addon/components/o-s-s/feature-cards-container.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export default class OSSFeatureCardsContainer extends Component<OSSFeatureCardsContainerArgs> {
+  constructor(owner: unknown, args: OSSFeatureCardsContainerArgs) {
+    super(owner, args);
+
+    assert('[OSS::FeatureCardsContainer] The @cards parameter is mandatory', Array.isArray(args.cards));
+    assert(
+      '[OSS::FeatureCardsContainer] @cards must contain exactly 2 or 3 cards',
+      args.cards.length === 2 || args.cards.length === 3
+    );
```
Miexil, line 38: Not sure we shouldn't also handle the `1` card scenario.
If we ever need some dynamic contents in the parents we'll be bothered by this - imo a single card just displays the single card centered in the container.

app/styles/molecules/feature-card.less
```diff
@@ -0,0 +1,64 @@
...
+    font-size: var(--font-size-sm);
+    color: var(--color-gray-600);
+    text-wrap: balance;
+    margin: 0;
+  }
+
+  &__illustration {
+    inline-size: 100%;
+    block-size: auto;
+    max-block-size: 123px;
```
Miexil: `123px` exactly ? 🫠 
Is that coming from the designs directly ?

tests/integration/components/o-s-s/feature-card-test.ts
```diff
@@ -0,0 +1,75 @@
...
+  });
+
+  test('it supports color and shadow variants', async function (assert) {
+    await render(
+      hbs`<OSS::FeatureCard @title={{this.title}} @description={{this.description}} @image={{this.image}} @colorVariant="blue" @shadowVariant="lg" />`
+    );
+
+    assert.dom('.oss-feature-card').hasClass('oss-feature-card--color-blue');
+    assert.dom('.oss-feature-card').hasClass('oss-feature-card--shadow-lg');
+  });
```
Miexil: Please write these as separate modules and checking all possible colors & shadows in loops 🙏

tests/integration/components/o-s-s/feature-card-test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  module('Error management', () => {
```
Miexil, line 85: It also throws for unknown shadow & color type if i'm not mistaken

addon/components/o-s-s/feature-card.ts
edouardmisset: From my understanding, if that happens we fail one of the `assert`s above (L. 42, L. 48). It seems to be the way other components do it (see empty-state or dialog components).

addon/components/o-s-s/feature-cards-container.ts
edouardmisset: The idea was to make this a object parameter. If later, the designers wanted to add a 4th card, we could simply add index `4` and define its structure there.
To answer more precisely, the number keys are the corresponding layout (2 => 2 cards, 3 => 3 cards, etc.). See line 76 `const layout = CARDS_LAYOUT[cards.length as keyof typeof CARDS_LAYOUT]`
I admit it might be a bit over-engineered. 
If you would prefer another structure let me know and I'll see what I can do. :)

app/styles/molecules/feature-card.less
edouardmisset: It is defined just underneath on lines 16, 21, 26.

app/styles/molecules/feature-card.less
edouardmisset: Figma uses `160%` if memory serves. In the web, I've usually seen it as a dimensionless number (`1`, `1.5`, etc.)
If you prefer I can set it like in Figma. 
For reference, [both are accepted](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/line-height).

app/styles/molecules/feature-card.less
edouardmisset: I just checked, it's not the default for `<h3>`s apparently.

app/styles/molecules/feature-card.less
edouardmisset: Yes, I just [double checked](https://www.figma.com/design/xd32CWO1lHvIfbJ7ZTZJxq/%E2%9A%99%EF%B8%8F-Provide-OAuth-Client-Credentials-Authentication-for-API-Customers?node-id=6304-44946&t=BeOqA5Om2Mpg71lj-4) the designs I based my work on are all `123px`.  
If you prefer I made the sizing more responsive (using relative units or another technique), I can, just let me know. I just tried to match the designs that's all.

addon/components/o-s-s/feature-card.hbs
edouardmisset: Not sure what you mean.
In the Dummy app, the h3 text is rendered at 14px like in the designs

image

I'm not sure if I'm missing something, if so what?

tests/dummy/app/controllers/visual.ts
edouardmisset: You're right, thank you.

addon/components/o-s-s/feature-card.hbs
Miexil: In `oss` we currently have `1rem` equal to

Image

The "relative" nature of rem makes me doubt that we'll always have a 14px where required.

I guess my point here is why are we not using `font-size-lg` and `font-size-md` where needed ?

addon/components/o-s-s/feature-cards-container.ts
Miexil: While we are now handling the single card use case (thank you 🙏), we still have this component dictate shadows etc

I know we often use config-based approaches in our WoW, but here it seems to me like the code would be better off with a couple of helpers for the rotation and the offset ?

app/styles/molecules/feature-card.less
Miexil: let's at least assign a more "round/pair" value with `122px` ?

app/styles/molecules/feature-card.less
phndiaye: This is actually the default on the platform so i think you can skip it:

- https://github.com/upfluence/oss-components/blob/8e387f9a3a3f315373aec36c66c53196e02629a2/app/styles/core/_typography.less#L16
- https://github.com/upfluence/oss-components/blob/8e387f9a3a3f315373aec36c66c53196e02629a2/app/styles/deprecated/_typography.less#L39 /

edouardmisset: Apparently the `h3` header is set to `1.1` in `type.less`

image

app/styles/molecules/feature-card.less
edouardmisset: Damn auto-complete. 
Thank you for the heads up.
What's weird is I check the font in the browser and it was `Inter`... It must have inherited it. Should I still set it or leave it inherited?

addon/components/o-s-s/feature-card.ts
edouardmisset: Thanks.

addon/components/o-s-s/feature-cards-container.hbs
```diff
@@ -0,0 +1,13 @@
+<div class="oss-feature-cards-container" ...attributes>
+  {{#each this.cardsWithLayout as |card|}}
+    <div class={{concat "oss-feature-cards-container__item " card.className}} style={{card.style}}>
```
aprentout: Do we really need this div ? 
Couldn't this be handled directly by the `...attributes` of the card directly ?

addon/components/o-s-s/feature-card.hbs
edouardmisset: Indeed the `oss-feature-card__title` class has the following definition:

`font-size: var(--font-size-md);`

I'm not sure what I'm missing.

Miexil: ah yes, indeed there's the `oss-feature-card__title` class which defines font-size, but then also the `<h3>` tag which defines another size

Why not just have:
 ```<span class="font-size-md">{{@title}}</span>``` 
no dedicated class which does not look useful, + no usage of `h3` tag which has its own font-size management

addon/components/o-s-s/feature-card.ts
phndiaye: I guess `SafeString` properly inherit string and `SafeString#trim` is a thing, but might be worth adding a test for it

addon/components/o-s-s/feature-cards-container.ts
```diff
@@ -0,0 +1,96 @@
...
+
+      const { colorVariant, shadowVariant, isCenter, rotation, offsetX } = cardLayout;
+
+      return {
+        ...card,
+        className: isCenter ? `${BASE_ITEM_CLASS} ${BASE_ITEM_CLASS}--center` : BASE_ITEM_CLASS,
+        colorVariant,
+        shadowVariant,
+        isCenter,
+        style: `transform: translateX(${offsetX}) rotate(${rotation}deg);`
```
phndiaye: `style: htmlSafe(...)` as it's later passed on an inline style (Ember only whines about it at rendering in the console ^^)

addon/components/o-s-s/feature-cards-container.ts
phndiaye: Disclaimer: you know my affinity w/ CSS 😅
But there is no way this is done w/ pure CSS ?

edouardmisset: Implemented the layout logic in CSS as proposed by @phndiaye and created a couple of helpers to set fallback colors and shadows. @Miexil

OwenCoogan (approved):

aprentout (approved):

JulienVannier66 (approved):

PR #669, opened 2026-05-20
add autocomplete arg to passwordInput

Elodie-DeMatteis-Upf:
### What does this PR do?
`OSS::PasswordInput` now supports a new `autocomplete` optional arg, being set either to `on` or `off`. It allows the input to disable its default autocompletion.

Related to: #

With `autocomplete` undefined
Capture d’écran 2026-05-20 à 11 24 08

With `autocomplete="off"`
Capture d’écran 2026-05-20 à 11 24 17

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [ ] Properly labeled

### DIFF
--- a/addon/components/o-s-s/password-input.hbs
+++ b/addon/components/o-s-s/password-input.hbs
@@ -7,7 +7,7 @@
         disabled={{@disabled}}
         placeholder={{this.placeholder}}
         autocomplete={{this.autocomplete}}
-        {{on "keyup" this.validateInput}}
+        {{on "input" this.validateInput}}
       />
     </:input>
     <:suffix>
--- a/addon/components/o-s-s/password-input.hbs
+++ b/addon/components/o-s-s/password-input.hbs
@@ -6,7 +6,7 @@
         @type={{this.visibility}}
         disabled={{@disabled}}
         placeholder={{this.placeholder}}
-        autocomplete="current-password"
+        autocomplete={{this.autocomplete}}
         {{on "keyup" this.validateInput}}
       />
     </:input>
--- a/addon/components/o-s-s/password-input.stories.js
+++ b/addon/components/o-s-s/password-input.stories.js
@@ -46,12 +46,13 @@ export default {
       control: { type: 'object' }
     },
     autocomplete: {
-      description: 'Whether or not the input should have autocomplete enabled.',
+      description:
+        'Whether or not the input should have autocomplete enabled. If set to "off", it will use "new-password" to prevent browsers from filling in the password. It can be set to any custom value if needed.',
       table: {
         type: {
-          summary: 'on | off'
+          summary: 'on | off | string'
         },
-        defaultValue: { summary: 'undefined' }
+        defaultValue: { summary: 'current-password' }
       },
       control: { type: 'text' }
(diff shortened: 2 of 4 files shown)


addon/components/o-s-s/password-input.hbs
```diff
@@ -6,7 +6,7 @@
         @type={{this.visibility}}
         disabled={{@disabled}}
         placeholder={{this.placeholder}}
-        autocomplete="current-password"
+        autocomplete={{this.autocomplete}}
         {{on "keyup" this.validateInput}}
```
aprentout: Could we replace it with "input" in order to handle validation on browser password autosuggest
```suggestion
        {{on "input" this.validateInput}}
```

Elodie-DeMatteis-Upf: Oh yes, forgot to update that, thanks!!

addon/components/o-s-s/password-input.stories.js
```diff
@@ -45,6 +45,16 @@ export default {
       },
       control: { type: 'object' }
     },
+    autocomplete: {
+      description: 'Whether or not the input should have autocomplete enabled.',
+      table: {
+        type: {
+          summary: 'on | off'
+        },
+        defaultValue: { summary: 'undefined' }
```
Miexil: defaults to `current-password`

addon/components/o-s-s/password-input.ts
```diff
@@ -20,6 +20,7 @@ interface OSSPasswordInputArgs {
   errorMessage?: string;
   feedbackMessage?: FeedbackMessage;
   disabled?: boolean;
+  autocomplete?: 'on' | 'off';
```
Miexil: From the MDN:
> The attribute value is either the keyword off or on, or an ordered list of space-separated tokens.

We also had `current-password` which is now the default `autocomplete` getter value, and we are setting to `new-password` when it's `off` ? 🤔 

There's a typing issue for me here

Miexil (commented):
And missing tests :/

addon/components/o-s-s/password-input.ts
phndiaye: @Miexil the `new-password` thing was discovered a couple sprints ago for EmailInput component, i remember it's because Chrome doesn't really support `off` but `new-password` is somehow a value that all browsers settled on for some reason (don't ask me why 😂)

Miexil: I'm not really questioning the `new-password` only mentioning that possible values are not only limited to `on` or `off` and taking the code from the getter below as an example

Elodie-DeMatteis-Upf: Yes, "current-password" will autocomplete the password field to your... well... current password. "new-password" states that the password that will be added is new, so there should not be any autocompletion on this field.
So to recap:
- Previous behavior was defaulting to "current-password" - that is still the case
- Adding `@autocomplete="off"` will use "new-password" to disable the autocomplete

Miexil (approved):

olxmpe (approved):

PR #675, opened 2026-05-26
data-control-name support in OSS::ToggleButtons opts

phndiaye:
### What does this PR do?

data-control-name support in OSS::ToggleButtons opts (and misc fixes that slipped in recently)

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled


addon/components/o-s-s/context-menu/panel.hbs
phndiaye, line 1: These are failing in 4.12 compat checks because it's more severe about it. Let's be careful about it, i believe the Linter actually whines about it in the editors 🤔  (but it's not linted in the CI)

addon/components/o-s-s/toggle-buttons.hbs
```diff
@@ -6,6 +6,7 @@
     <div
       class="oss-toggle-buttons-btn {{if (eq @selectedToggle toggle.value) ' oss-toggle-buttons-btn--selected'}}"
       role="button"
+      data-control-name={{concat "oss-toggle-button-option-" toggle.value}}
```
phndiaye, line 9: Main topic of the PR ^^

tests/integration/components/o-s-s/smart/immersive/currency-input-test.ts
```diff
@@ -205,7 +172,7 @@ module('Integration | Component | o-s-s/smart/immersive/currency-input', functio
     });
   });
 
-  module('Loading', () => {
+  module('Loading', (hooks) => {
```
phndiaye, line 175: the `hooks` below was relying on the wrong scope, which can lead to flaky tests

addon/utils/logo-config.ts
phndiaye, line 1: app re-export files should not be in `.ts` and "implems" shouldn't be in `app` dir

nathalieArnoux (approved):

OwenCoogan (approved):

JulienVannier66 (approved):

PR #676, opened 2026-05-28
Handle vertical overflow in enable-tooltip

olxmpe:
### What does this PR do?

Capture d’écran 2026-05-28 à 11 23 39

Related to: #

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [ ] Properly labeled

### DIFF
--- a/addon/utils/element.ts
+++ b/addon/utils/element.ts
@@ -2,12 +2,9 @@ export type OverflowDirection = 'horizontal' | 'vertical';
 
 export function hasOverflow(element: HTMLElement, direction: OverflowDirection = 'horizontal'): boolean {
   if (!Array.from(element.children).length) {
-    switch (direction) {
-      case 'horizontal':
-        return element.offsetWidth < element.scrollWidth;
-      case 'vertical':
-        return element.offsetHeight < element.scrollHeight;
-    }
+    return direction === 'horizontal'
+      ? element.offsetWidth < element.scrollWidth
+      : element.offsetHeight < element.scrollHeight;
   }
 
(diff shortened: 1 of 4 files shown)


addon/utils/element.ts
```diff
@@ -1,9 +1,16 @@
...
+
+export function hasOverflow(element: HTMLElement, direction: OverflowDirection = 'horizontal'): boolean {
   if (!Array.from(element.children).length) {
-    return element.offsetWidth < element.scrollWidth;
+    switch (direction) {
+      case 'horizontal':
+        return element.offsetWidth < element.scrollWidth;
+      case 'vertical':
+        return element.offsetHeight < element.scrollHeight;
+    }
```
aprentout: Nit: Open question here but as there is only two cases, do we really need a switch ?


```suggestion
  if(direction === 'horizontal') return element.offsetHeight < element.scrollHeight;
  return element.offsetWidth < element.scrollWidth;
```

aprentout (approved):

addon/utils/element.ts
olxmpe: I hesitated tu put "both" as an option but I don't think it really makes sense for the moment, so indeed not really needed

phndiaye: @aprentout your suggestion flipped height & width
@olxmpe be careful about that ^^
otherwise yes i think the switch isn't worth it here

JulienVannier66 (approved):

phndiaye (approved):

PR #678, opened 2026-06-08
Updated OSS::Alert component

nathalieArnoux:
### What does this PR do?

Updated `OSS::Alert` component to fit the new design.
Mainly removed colored indicator on the left, and modified colors.
Updated tests.

Related to: #[dra-5252](https://linear.app/upfluence/issue/DRA-5252/groundworkoss-components-update-alert-component)

Capture d’écran 2026-06-10 à 10 07 34

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/addon/components/o-s-s/alert.hbs
+++ b/addon/components/o-s-s/alert.hbs
@@ -1,24 +1,25 @@
-<div class={{concat "upf-alert " this.skinClass}} {{did-insert this.initSelf}} ...attributes>
-  <div class="indicator"></div>
-  <div class="main-container {{if this.plain 'main-container--plain'}}">
-    <span class="icon">
-      <OSS::Icon @icon={{this.iconClass}} />
-    </span>
-    <div class="text-container">
-      {{#if @title}}
-        <span class="title">{{@title}}</span>
-      {{/if}}
-      {{#if @subtitle}}
-        <span class="subtitle">{{@subtitle}}</span>
-      {{/if}}
-      {{#if (has-block "extra-content")}}
-        {{yield to="extra-content"}}
-      {{/if}}
-    </div>
-    {{#if @closable}}
-      <div class="fx-col">
-        <i role="button" class="far fa-times" {{on "click" this.removeSelf}}></i>
-      </div>
+<div
+  class={{concat "upf-alert " this.skinClass (if this.plain " upf-alert--plain")}}
+  {{did-insert this.initSelf}}
+  ...attributes
+>
+  <span class="icon">
+    <OSS::Icon @icon={{this.iconClass}} />
+  </span>
+  <div class="text-container">
+    {{#if @title}}
+      <span class="title">{{@title}}</span>
+    {{/if}}
+    {{#if @subtitle}}
+      <span class="subtitle">{{@subtitle}}</span>
+    {{/if}}
+    {{#if (has-block "extra-content")}}
+      {{yield to="extra-content"}}
     {{/if}}
   </div>
+  {{#if @closable}}
+    <div class="fx-col">
+      <i role="button" class="far fa-times" {{on "click" this.removeSelf}}></i>
+    </div>
+  {{/if}}
 </div>
--- a/addon/components/o-s-s/alert.hbs
+++ b/addon/components/o-s-s/alert.hbs
@@ -14,15 +14,13 @@
     {{/if}}
   </div>
   {{#if @closable}}
-    <div class="fx-col">
-      <OSS::Button
-        class="upf-alert-close-button"
-        @size="sm"
-        @icon="fa-times"
-        @square={{true}}
-        {{on "click" this.removeSelf}}
-        data-control-name="upf-alert-close-button"
-      />
-    </div>
+    <OSS::Button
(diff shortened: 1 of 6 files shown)


addon/components/o-s-s/alert.hbs
```diff
@@ -1,24 +1,25 @@
...
-      {{#if (has-block "extra-content")}}
-        {{yield to="extra-content"}}
-      {{/if}}
-    </div>
-    {{#if @closable}}
-      <div class="fx-col">
-        <i role="button" class="far fa-times" {{on "click" this.removeSelf}}></i>
-      </div>
+<div
+  class={{concat "upf-alert " this.skinClass (if this.plain " upf-alert--plain")}}
```
olxmpe: Why don't you add the --plain from the skinClass getter?

addon/components/o-s-s/alert.stories.js
```diff
@@ -93,15 +94,15 @@ const defaultArgs = {
...
                   @closable={{this.closable}} />
   `,
   context: args
 });
 
 const ExtraContentTemplate = (args) => ({
   template: hbs`
-      <OSS::Alert @skin={{this.skin}} @title={{this.title}} @subtitle={{this.subtitle}} @plain={{this.plain}} 
+      <OSS::Alert @skin={{this.skin}} @title={{this.title}} @subtitle={{this.subtitle}} @plain={{this.plain}}
                   @closable={{this.closable}}>
```
edouardmisset: question: don't you need to pass the `onClose` handler defined L. 62?

olxmpe (approved):

addon/components/o-s-s/alert.hbs
```diff
@@ -1,24 +1,25 @@
...
+      <span class="subtitle">{{@subtitle}}</span>
+    {{/if}}
+    {{#if (has-block "extra-content")}}
+      {{yield to="extra-content"}}
     {{/if}}
   </div>
-</div>
+  {{#if @closable}}
+    <div class="fx-col">
+      <i role="button" class="far fa-times" {{on "click" this.removeSelf}}></i>
```
olxmpe: Would oss::icon work here instead of i?

app/styles/alert.less
```diff
@@ -5,81 +5,93 @@
...
-  }
+  padding: var(--spacing-px-9);
+  gap: var(--spacing-px-12);
+  background-color: var(--color-gray-50);
 
   .text-container {
     display: flex;
     flex-direction: column;
     flex: 1;
     gap: 3px;
```
edouardmisset, line 16: question: I don't see this `3px` gap in the designs. Am I missing something? Where does it come from?

addon/components/o-s-s/alert.hbs
JulienVannier66:
```suggestion
  class={{concat "upf-alert" this.skinClass (if this.plain " upf-alert--plain")}}
```
You already put the space in the second class

addon/components/o-s-s/alert.hbs
nathalieArnoux: Actually it was defined in the designs that it should be a transparent button, I completely let this slip though so thanks for your comment! :D

app/styles/alert.less
nathalieArnoux: This is the gap between the title and subtitle, it is indeed defined in the designs, and it hasn't changed from the first design :)

Image

addon/components/o-s-s/alert.stories.js
nathalieArnoux: I added them but I'm not sure it was necessary since `closable` is defined by default as `false` 🤔 
Not sure though, so if anyone can confirm I'd love to know :)

Miexil: Yep, it's required if we want to be able to update the closable value in storybook directly and have it reflect in storybook's UI

app/styles/alert.less
```diff
@@ -5,81 +5,98 @@
...
       }
     }
   }
 
-  .indicator {
-    width: 3px;
+  .upf-square-btn.upf-btn--default.upf-square-btn--sm {
+    background-color: transparent;
+    border: none;
   }
```
Miexil: I'm afraid if we do it this way we might have these attribute updates leaking on other buttons when using the "extra-content" named-block.

You can double check that locally, but I suggest you append a specific class on the button itself in the template in which you can assign these CSS attributes.

tests/integration/components/o-s-s/alert-test.ts
```diff
@@ -42,22 +42,22 @@ module('Integration | Component | o-s-s/alert', function (hooks) {
   });
```
Miexil, line 43: could you please add an assertion that matches the correct class here ?

Miexil (approved):

edouardmisset (approved):

Miexil (approved):

Miexil (approved):

aprentout (approved):

JulienVannier66 (approved):

PR #679, opened 2026-06-09
Stop silently pushing assets into consumer apps

phndiaye:
### What does this PR do?

Till now, this addon has been pushing assets into consumer apps' public tree. This is an anti-pattern as it leads to weird situations where:

- people don't know where those assets come from
- people can remove the assets from the package if they don't see it being used locally, leading to missing assets in consumer apps later on
- it feels like magic but only people with the context know what's happening

This pattern is something that's now advised against and prevents moving to more modern package builds (Embroider, v2 addons) which rely on everything being explicit.

⚠️ This PR is to be merged only once affected consumer apps have properly stopped relying on the pushed assets.

Multiple changes occur here:

- Prefix assets path with `/@upfluence/oss-components/`: this is to adapt to the fact that assets are not merged with consumer apps anymore and local assets must be properly addressed (cf. `oss/smart/immersive/logo` file)
- `htmlSafe` return type: this is to fix a warning appearing whenever we have inline styles as it has to be htmlSafe
- `upf-icons` removal: upf-icons is an attempt from around 2017/2018 to have our own custom fonts. We stopped that pretty fast and there are no occurences of it anymore in the codebase
- in `index.js` we stop pushing assets into consumer apps
- cleanup unused files
- assets that are now only used in the demo app have been moved to `tests/dummy/public` to not pollute the package's assets.
- cleaned up assets paths related to upf-icon have been replaced in Stories with remaining ones

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled


JulienVannier66 (approved):

addon/components/o-s-s/upload-item.hbs
```diff
@@ -22,8 +22,7 @@
 
     <div class="fx-row fx-gap-px-6 fx-xalign-center">
       {{#if this.uploading}}
-        {{! template-lint-disable no-triple-curlies}}
-        <div class="oss-upload-item__loader" style={{{this.loaderStyle}}}>
+        <div class="oss-upload-item__loader" style={{this.loaderStyle}}>
```
Miexil, line 25: no need to specify it's a safe string or something ?

Miexil: Ah nevermind, was in the next file ^^

Miexil (approved):

PR #680, opened 2026-06-09
Introduce new text-carousel modifier

edouardmisset:
### What does this PR do?

Adds a `text-carousel` modifier that displays a list of labels over time using a configurable interval. The modifier displays the first label immediately, advances through each provided label once, and holds the final label until the element is updated or destroyed.

Related to: [#DRA-5309](https://linear.app/upfluence/issue/DRA-5309/groundworkoss-componentsember-identity-text-carousel-modifier-and)

- Consumers can use `{{text-carousel this.labels intervalMs=3000}}` to display rotating text in an element.
- Empty label lists clear the target element text.

Spec:
image

Implementation:
image

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [ ] Added/updated documentation with Figma design link. Don't forget to replace "design" by "file" in the URL. For example <https://www.figma.com/file/example>
- [x] Properly labeled


addon/modifiers/text-carousel.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  modify(
+    element: HTMLElement,
+    [labels]: TextCarouselArgs['positional'],
+    { intervalMs = 3000 }: TextCarouselArgs['named']
+): void {
+    this.clearInterval();
+
+    if (!labels?.length) {
+      element.textContent = '';
```
Miexil: As we are in a shared library, maybe it would make sense here to throw instead ?
@phndiaye what's your take ?

addon/modifiers/text-carousel.ts
Miexil, line 1: I think we should also update the modifiers.stories.mdx file with a new line for this one 🙏

addon/modifiers/text-carousel.ts
```diff
@@ -0,0 +1,52 @@
+import Modifier from 'ember-modifier';
+import { registerDestructor } from '@ember/destroyable';
+
+interface TextCarouselArgs {
+  positional: [labels?: string[]];
+  named: {
+    intervalMs?: number;
```
phndiaye: nit: just `interval` ? (but for ms is the default unit for everything, maybe that's not the case for everyone ^^)

addon/modifiers/text-carousel.ts
phndiaye: +1 for throwing if no labels are passed

addon/modifiers/text-carousel.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    if (labels.length === 1) return;
+
+    this.intervalId = window.setInterval(() => {
+      index++;
+      element.textContent = labels[index] ?? '';
+
+      if (index >= labels.length - 1) {
+        this.clearInterval();
+      }
```
phndiaye: what about having a `loop` (default: false) param to conditionally stop the rotation or keep it running?

addon/modifiers/text-carousel.ts
```diff
@@ -0,0 +1,52 @@
...
+      element.textContent = '';
+      return;
+    }
+
+    let index = 0;
+    element.textContent = labels[index] ?? '';
+
+    if (labels.length === 1) return;
+
+    this.intervalId = window.setInterval(() => {
```
phndiaye, line 36: should there also be a small transition or in your case we just have the text changing and that's it?

just in case we handled it this way in the search : https://github.com/upfluence/upfluence-web/blob/54461f972598c1c47277f4731d2638b80741bff2/packages/facade-web/addon/components/find-creators/query-builder/prompt-rules.ts#L132-L137.

But as this is generic, it would mean having the modifier set an inline css for `transition` rule 🤔

addon/modifiers/text-carousel.ts
JulienVannier66: +1 for Max comment

tests/integration/modifiers/text-carousel-test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+      clock.tick(30);
+      assert.dcn('carousel').hasText('Putting it together...');
+
+      clock.tick(6_000);
+      assert.dcn('carousel').hasText('Putting it together...');
+    } finally {
+      clock.restore();
+    }
+  });
```
JulienVannier66, line 38: Why not declare the clock in the `beforeEach` and do the restore in `afterEach`. More clean not, to avoid flaky effect

addon/modifiers/Modifiers.stories.mdx
```diff
@@ -27,3 +27,4 @@ According to Ember JS:
 | on-click-outside       | Triggers the passed callback function when a user clicks outside the element               | `<div {{on-click-outside callbackFunction}}>Floating Div</div>` |                             |
 | on-bottom-reached      | Triggers the passed callback function when a user reaches the bottom of the target element | `<div {{on-bottom-reached callbackFunction}}>Div</div>`         |                             |
 | scroll-shadow          | Add bottom shadow for scrollable element                                                   | `<div {{scroll-shadow color="field"}}>`                         | color: "default" (white)    |
+| text-carousel          | Displays each provided label over time and keeps the last one visible by default            | `<div {{text-carousel this.labels intervalMs=3000 loop=true}}></div>` | intervalMs: 3000, loop: false |
```
Miexil: `intervalMs` was changed to `invertval` if I'm not mistaken

addon/modifiers/text-carousel.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  modify(
+    element: HTMLElement,
+    [labels]: TextCarouselArgs['positional'],
+    { interval = 3000, loop = false }: TextCarouselArgs['named']
+  ): void {
+    this.clearInterval();
+
+    if (!labels?.length) {
+      throw new Error('TextCarouselModifier requires a non-empty labels array');
```
Miexil: In the BaseDropdown and Carousel components we have this kind of format : 
> [component][OSS::BaseDropdown] You must implement handleSelectorClose method on the child class

maybe set `[modifier][text-carousel] Requires a non-empty labels array` ?

tests/integration/modifiers/text-carousel-test.ts
```diff
@@ -0,0 +1,73 @@
...
+    this.labels = ['Thinking...', 'Analyzing your request...', 'Putting it together...'];
+    this.clock = sinon.useFakeTimers();
+  });
+
+  hooks.afterEach(function () {
+    this.clock.restore();
+  });
+
+  test('it displays the first label immediately', async function (assert) {
+    await render(hbs`<div data-control-name="carousel" {{text-carousel this.labels intervalMs=30}}></div>`);
```
Miexil: `intervalMs` was changed to `interval` I believe this also explains the failing tests in the CI

Miexil: Let's also set the base branch to master so that you merge in there versus in the feature branch 🙏

addon/modifiers/text-carousel.ts
edouardmisset: I'm entirely sure about that as we don't know what html element the modifier is being applied to.
You solution involve JS to make a fade-out/fade-in. To achieve this I would prefer a "simple" CSS solution.

Idk. The whole thing is a little awkward to me as we introduce a modifier here but another new component uses it in `ember-influencer`. We are coupling things (like the timing) in different repos... 

I'm a little lost to be honest. It all seems like a foot-gun now that I think of it.

JulienVannier66 (approved):
LGTM

aprentout (approved):

PR #683, opened 2026-06-12
Networkless font loading

phndiaye:
### What does this PR do?

This PR gets rid of CDN-loaded fonts (via Google web fonts)
- Get rid of legacy GT-Eesti font files (unused for years now)
- Stop loading fonts through CDN

Additionally, this is ditching `enable-popover` and `enable-dropdown` modifiers which were not working anyway as they rely on bootstrap's JS which we removed months ago.

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled


Miexil (approved):

OwenCoogan (approved):

app/styles/core/_font-faces.less
```diff
@@ -0,0 +1,313 @@
+/* inter-latin-ext-400-normal */
+@font-face {
```
edouardmisset, line 1: suggestion: It might be interesting to use variable font here so that we only serve one font instead of n variations

edouardmisset (approved):

JulienVannier66 (approved):

Miexil (approved):

OwenCoogan (approved):

PR #688, opened 2026-06-29
Avoid class duplication for text-size-x

JulienVannier66:
### What does this PR do?

Avoid class duplication for text-size-x

New compiled less file size:
2.1M packages/upfluence-web/dist/assets/upfluence-web.css 84 166 lignes (without fix)
1.9M packages/upfluence-web/dist/assets/upfluence-web.css 75 344 lignes (upfluence-web fix)
1.6M packages/upfluence-web/dist/assets/upfluence-web.css 51 276 lignes (upfluence-web + oss fixes)

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [ ] Added/updated tests
- [x] Properly labeled

### DIFF
--- a/app/styles/shims/_all.less
+++ b/app/styles/shims/_all.less
@@ -113,18 +113,45 @@
 // Text Size Utility classes
 // --------------------------------------------------
 
-.text-size(@id, @size) {
-  .text-size-@{id} {
-    font-size: @size / @rem;
-  }
+.text-size-1 {
+  font-size: 8 / @rem;
 }
 
-@text-sizes: 8, 10, 12, 14, 16, 18, 24, 30, 36, 42;
+.text-size-2 {
+  font-size: 10 / @rem;
+}
 
-.for(@text-sizes, {
-  .text-size(@i, @value);
-});
-.text-size(normal, @font-size-body);
+.text-size-3 {
+  font-size: 12 / @rem;
+}
+
+.text-size-4 {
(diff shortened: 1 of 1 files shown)


app/styles/shims/_all.less
phndiaye, line 1: for the number of classes generated, can't we just hardcode : `text-size-X` and drop the iterations? (those are deprecated classes anyway)

app/styles/shims/_all.less
```diff
@@ -113,29 +113,45 @@
...
+  font-size: 30 / @rem;
+}
+
+.text-size-9 {
+  font-size: 36 / @rem;
+}
+
+.text-size-10 {
+  font-size: 42 / @rem;
+}
```
edouardmisset, line 154: question: Why not use pure css? That would be one (n) less less (🤣) transformation and would bring us closer to pure css standard.
Here is an example:

```suggestion
text-size-10 {
  font-size: calc(42 / 12 * 1rem);
}
```

If I remember correctly our base font-size unit is `12`, right?

JulienVannier66: Yes but this class is a deprecated one, I don't want to change the value to don't break the soft

edouardmisset (approved):

Miexil (approved):

aprentout (approved):

PR #689, opened 2026-07-03
Smart::Blob - introduce new `xs` size

Miexil:
### What does this PR do?

Smart::Blob - introduce new `xs` size

Screenshot 2026-07-03 at 17 48 53

### Good PR checklist
- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [x] Properly labeled


phndiaye (approved):

addon/components/o-s-s/smart/blob.stories.js
```diff
@@ -17,13 +17,13 @@ export default {
...
       table: {
-        type: { summary: 'sm | md | xl' },
+        type: { summary: 'xs | sm | md | xl' },
         defaultValue: { summary: 'md' }
       },
       control: {
         type: 'select'
       },
-      options: ['sm', 'md', 'xl']
+      options: ['xs', 'sm', 'md', 'xl']
```
edouardmisset, line 26: nitnit: could be DRYer

phndiaye: @edouardmisset if you mean importing the array from the implem, Storybook can't :p  (Ember hell)

edouardmisset (approved):

nathalieArnoux (approved):

addon/components/o-s-s/smart/blob.stories.js
edouardmisset: yeah I know.
Actually I meant between L. 20 and L. 26. To have a single source of truth (at least in this file).
But it might not be worth it honestly

aprentout (approved):

PR #697, opened 2026-08-04
Groundwork: Add OSS::TogglableSection option to remove the switch

olxmpe:
### What does this PR do?

Add an option in togglable-section component to not display the toggle switch, making sure that it does not impact current behavior

Capture d’écran 2026-08-04 à 16 59 30

Related to: #

### Good PR checklist

- [x] Title makes sense
- [x] Properly assigned
- [x] Added/updated tests
- [ ] Properly labeled

### DIFF
--- a/addon/components/o-s-s/togglable-section.ts
+++ b/addon/components/o-s-s/togglable-section.ts
@@ -29,7 +29,7 @@ export default class CampaignTogglableSection extends Component<CampaignTogglabl
   }
 
   get isSwitchable(): boolean {
-    return this.args.switchable !== false;
+    return this.args.switchable ?? true;
   }
 
   @action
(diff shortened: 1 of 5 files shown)


OwenCoogan (approved):

nathalieArnoux (approved):

addon/components/o-s-s/togglable-section.ts
```diff
@@ -27,9 +28,13 @@ export default class CampaignTogglableSection extends Component<CampaignTogglabl
     return this.args.size === 'sm' ? 'padding-px-12' : 'padding-px-18';
   }
 
+  get isSwitchable(): boolean {
+    return this.args.switchable !== false;
```
JulienVannier66:
```suggestion
    return this.args.switchable ?? true;
```
You can use nullish coalescing

JulienVannier66 (approved):
Otherwise LGTM

PR #700, opened 2026-08-19
OSS::Stats::Banner

OwenCoogan:
### What does this PR do?

Screenshot 2026-08-19 at 17 17 31

Introduces a new OSS::Stats::Banner component for flexible KPI/header composition.

* Adds structured config args for a few elements ( badge & icon )
* Uses dropdown, extra-badges, and cta as named blocks to preserve flexibility in those sections.
* Implements title ellipsis with a top-position tooltip shown only on overflow.
* Renders the info icon when titleInfoCircle is provided.
* Hides extra-badges and cta block content in loading mode for consistent loading UI.
* Updates to Storybook and tests

Related to: #

### Good PR checklist

- [X] Title makes sense
- [X] Is against the correct branch
- [X] Only addresses one issue
- [X] Properly assigned
- [X] Added/updated tests
- [X] Added/updated documentation with Figma design link. Don't forget to replace "design" by "file" in the URL. For example [https://www.figma.com/file/example](<https://www.figma.com/file/example>)
- [X] Properly labeled

### DIFF
--- a/addon/components/o-s-s/stats/banner.hbs
+++ b/addon/components/o-s-s/stats/banner.hbs
@@ -1,0 +1,63 @@
+<div class="oss-stats-banner fx-col fx-gap-px-12" ...attributes>
+  <div class="oss-stats-banner__header fx-row fx-gap-px-12 fx-malign-space-between">
+    <div class="oss-stats-banner__header_content fx-row fx-gap-px-12 fx-malign-center fx-xalign-center">
+      {{#if (and this.loading this.hasBadgeContent)}}
+        <OSS::Skeleton @height={{36}} @width={{36}} class="oss-stats-banner__skeleton--badge" />
+      {{else if this.hasBadgeContent}}
+        <OSS::Badge
+          @icon={{@badgeIcon}}
+          @image={{@badgeImage}}
+          @text={{@badgeText}}
+          @skin={{@badgeSkin}}
+          @size={{@badgeSize}}
+          @plain={{@badgePlain}}
+        />
+      {{/if}}
+      <div class="oss-stats-banner__content">
+        {{#if (has-block "content")}}
+          {{yield to="content"}}
+        {{else if (and this.loading @title)}}
+          <OSS::Skeleton @height={{18}} @width={{420}} class="oss-stats-banner__skeleton--title" />
+        {{else}}
+          <div class="header-section fx-row fx-gap-px-8 fx-malign-space-between">
+            <span class="font-color-gray-600">{{@title}}</span>
+          </div>
+        {{/if}}
+      </div>
+    </div>
+
+    <div>
+      {{#if (and this.loading (has-block "extra-badges"))}}
+        <div class="fx-row fx-gap-px-6">
+          <OSS::Skeleton @height={{28}} @width={{28}} class="oss-stats-banner__skeleton--extra-badge" />
+          <OSS::Skeleton @height={{28}} @width={{28}} class="oss-stats-banner__skeleton--extra-badge" />
+        </div>
+      {{else if (has-block "extra-badges")}}
+        {{yield to="extra-badges"}}
+      {{/if}}
+    </div>
+  </div>
+  <div class="oss-stats-banner__bottom_content fx-row fx-gap-px-12 fx-malign-space-between">
+    <div class="oss-stats-banner__stat fx-row fx-gap-px-4">
+      {{#if (and this.loading @statValue)}}
+        <OSS::Skeleton @height={{18}} @width={{90}} class="oss-stats-banner__skeleton--stat-value" />
+      {{else}}
+        <span class="font-color-gray-600">{{@statValue}}</span>
+
+      {{/if}}
+
+      {{#if (and this.loading @statExtraInfo)}}
+        <OSS::Skeleton @height={{16}} @width={{120}} class="oss-stats-banner__skeleton--stat-extra" />
+      {{else if @statExtraInfo}}
+        <span class="font-color-gray-400">{{@statExtraInfo}}</span>
+      {{/if}}
+    </div>
+    {{#if (and this.loading (has-block "cta"))}}
+      <OSS::Skeleton @height={{32}} @width={{120}} class="oss-stats-banner__skeleton--cta" />
+    {{else if (has-block "cta")}}
+      <div class="oss-stats-banner__cta">
+        {{yield to="cta"}}
+      </div>
+    {{/if}}
+  </div>
+</div>
--- a/addon/components/o-s-s/stats/banner.hbs
+++ b/addon/components/o-s-s/stats/banner.hbs
@@ -1,10 +1,17 @@
 <div class="oss-stats-banner fx-col fx-gap-px-12" ...attributes>
   <div class="oss-stats-banner__header fx-row fx-gap-px-12 fx-malign-space-between">
-    <div class="oss-stats-banner__header_content fx-row fx-gap-px-12 fx-malign-center fx-xalign-center">
-      {{#if (and this.loading (has-block "badges"))}}
-        <OSS::Skeleton @height={{36}} @width={{36}} class="oss-stats-banner__skeleton--badge" />
-      {{else if (has-block "badges")}}
-        {{yield to="badges"}}
+    <div class="oss-stats-banner__header_content fx-row fx-gap-px-9 fx-malign-center fx-xalign-center">
+      {{#if @badge}}
+        {{#if @extraIcon}}
+          <OSS::Icon @icon={{@extraIcon.icon}} style={{if @extraIcon.color (concat "color:" @extraIcon.color)}} />
+        {{/if}}
+        <OSS::Badge
+          @icon={{@badge.icon}}
+          @image={{@badge.image}}
+          @text={{@badge.text}}
+          @skin={{@badge.skin}}
+          @size="sm"
+        />
       {{/if}}
       <div class="oss-stats-banner__content">
         {{#if (has-block "content")}}
@@ -12,8 +19,16 @@
         {{else if (and this.loading @title)}}
           <OSS::Skeleton @height={{18}} @width={{420}} class="oss-stats-banner__skeleton--title" />
         {{else}}
-          <div class="header-section fx-row fx-gap-px-8 fx-malign-space-between">
-            <span class="font-color-gray-600">{{@title}}</span>
+          <div class="header-section fx-row fx-gap-px-8 fx-malign-space-between fx-xalign-center">
+            <span class="font-color-gray-600 font-weight-semibold">{{@title}}</span>
+            {{#if @titleInfo}}
+              <span {{enable-tooltip title=@titleInfo placement="top"}}>
+                <OSS::Icon @icon="fa-circle-info" />
+              </span>
+            {{/if}}
+            {{#if (has-block "dropdown")}}
+              {{yield to="dropdown"}}
+            {{/if}}
           </div>
         {{/if}}
       </div>
@@ -31,18 +46,25 @@
     </div>
   </div>
   <div class="oss-stats-banner__bottom_content fx-row fx-gap-px-12 fx-malign-space-between">
-    <div class="oss-stats-banner__stat fx-row fx-gap-px-4">
+    <div class="oss-stats-banner__stat fx-row fx-gap-px-6 fx-xalign-center">
       {{#if (and this.loading @statValue)}}
         <OSS::Skeleton @height={{18}} @width={{90}} class="oss-stats-banner__skeleton--stat-value" />
       {{else}}
-        <span class="font-color-gray-600">{{@statValue}}</span>
-
+        <span class="font-color-gray-900 font-weight-semibold font-size-h3">{{@statValue}}</span>
       {{/if}}
-
       {{#if (and this.loading @statExtraInfo)}}
         <OSS::Skeleton @height={{16}} @width={{120}} class="oss-stats-banner__skeleton--stat-extra" />
       {{else if @statExtraInfo}}
-        <span class="font-color-gray-400">{{@statExtraInfo}}</span>
+        <span class="font-color-gray-500">{{@statExtraInfo}}</span>
+      {{/if}}
+
+      {{#if @extraInfoTag}}
+        <OSS::Tag
+          @label={{@extraInfoTag.label}}
+          @skin={{@extraInfoTag.skin}}
+          @icon={{@extraInfoTag.icon}}
+          @plain={{@extraInfoTag.plain}}
+        />
       {{/if}}
     </div>
     {{#if (and this.loading (has-block "cta"))}}
--- a/addon/components/o-s-s/stats/banner.hbs
+++ b/addon/components/o-s-s/stats/banner.hbs
@@ -1,66 +1,74 @@
 <div class="oss-stats-banner fx-gap-px-12" ...attributes>
   <div class="oss-stats-banner__header fx-malign-space-between">
     <div class="oss-stats-banner__header_content fx-row fx-gap-px-9 fx-malign-center fx-xalign-center">
-      {{#if @badge}}
-        {{#if @extraIcon}}
-          <OSS::Icon @icon={{@extraIcon.icon}} style={{if @extraIcon.color (concat "color:" @extraIcon.color)}} />
+      {{#if @badgeConfig}}
+        {{#if @extraIconConfig}}
+          <OSS::Icon
+            @icon={{@extraIconConfig.icon}}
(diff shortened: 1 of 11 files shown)


addon/components/o-s-s/stats/banner.hbs
```diff
@@ -0,0 +1,75 @@
...
+            <span
+              class="oss-stats-banner__title text-ellipsis font-color-gray-600 font-weight-semibold"
+              {{enable-tooltip title=@title placement="top" displayOnlyOnOverflow=true}}
+            >{{@title}}</span>
+            <div class="oss-stats-banner__title-actions">
+              {{#if @titleInfoCircle}}
+                <span {{enable-tooltip title=@titleInfoCircle placement="top"}}>
+                  <OSS::Icon @icon="fa-info-circle" class="font-color-gray-400" />
+                </span>
+              {{/if}}
```
Miexil: I'm wondering if this one shouldn't be a named-block instead.

IMO it really depends on how *often* it is used. If we have it in at least 50% of the use-cases, might as well leave it this way; it'll be lighter on the parent caller

@phndiaye wdyt ?

addon/components/o-s-s/stats/banner.hbs
```diff
@@ -0,0 +1,75 @@
...
+      {{/if}}
+      <div class="oss-stats-banner__content">
+        {{#if @loading}}
+          <OSS::Skeleton @height={{12}} @width={{150}} class="oss-stats-banner__skeleton--title" />
+        {{else}}
+          <div class="oss-stats-banner__title-row">
+            <span
+              class="oss-stats-banner__title text-ellipsis font-color-gray-600 font-weight-semibold"
+              {{enable-tooltip title=@title placement="top" displayOnlyOnOverflow=true}}
+            >{{@title}}</span>
```
Miexil: tiniest nit, can we have the `{{@title}}` on its own line ? 
```
<span
  class="..."
  {{enable-tooltip ...}}
>
  {{@title}}
</span>
```

app/styles/organisms/stats-banner.less
```diff
@@ -0,0 +1,157 @@
...
+  &__skeleton--title {
+    max-width: 100%;
+  }
+
+  &__skeleton--cta {
+    border-radius: var(--border-radius-xl);
+  }
+}
+
+@container (width <= 900px) {
```
Miexil, line 181: ❤️

tests/integration/components/o-s-s/stats/banner-test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  test('when @extraIconConfig has a color, it applies the color as inline style', async function (assert) {
+    await render(hbs`
+      <OSS::Stats::Banner @badgeConfig={{hash icon="fa-check" skin="success"}} @extraIconConfig={{hash icon="fa-star" color="red"}} />
+    `);
+
+    assert.dom('.oss-stats-banner .fa-star').hasAttribute('style', 'color:red');
+  });
+
+  test('when @titleInfoCircle is defined, it renders an info-circle icon next to the title', async function (assert) {
```
Miexil: needs a counter test, ie, when not defined, not showing up

tests/integration/components/o-s-s/stats/banner-test.ts
```diff
@@ -0,0 +1,163 @@
...
+    await render(hbs`
+      <OSS::Stats::Banner @title="My stat" @titleInfoCircle="Helpful description" />
+    `);
+
+    assert.dom('.oss-stats-banner .fa-info-circle').exists();
+  });
+
+  test('when the dropdown block is defined, it renders its content in the title row', async function (assert) {
```
Miexil: same here, missing counter test

tests/integration/components/o-s-s/stats/banner-test.ts
```diff
@@ -0,0 +1,163 @@
...
+        <:dropdown>
+          <div class="test-dropdown">Dropdown</div>
+        </:dropdown>
+      </OSS::Stats::Banner>
+    `);
+
+    assert.dom('.oss-stats-banner .test-dropdown').exists();
+  });
+
+  test('when @loading is false and title/stat values are defined, it renders text instead of skeletons', async function (assert) {
```
Miexil: move to loading module

tests/integration/components/o-s-s/stats/banner-test.ts
```diff
@@ -0,0 +1,163 @@
+import { hbs } from 'ember-cli-htmlbars';
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+
+module('Integration | Component | o-s-s/stats/banner', function (hooks) {
```
Miexil, line 6: global comment here, let's move related topics into sub-module, for example:
- Module 'Loading state handling'
- - when `@loading` is true and no related args or blocks are defined, it renders title/stat skeletons
- - when `@loading` is true, it renders title/stat skeletons and hides named block content
- - when `@loading` is false and title/stat values are defined, it renders text instead of skeletons
- module for extraIconConfig

etc

Miexil (commented):
Don't have mutch to say here except in the tests, good work

addon/components/o-s-s/stats/banner.hbs
OwenCoogan: Yeah I was actually debating with myself yesterday if the info-circle & dropdown should both be wrapped in a title extra actions block.
For now in the redesign, we have 13 banners to update & only 2 contain info-circles 👀

addon/components/o-s-s/stats/banner.hbs
```diff
@@ -0,0 +1,77 @@
...
+          @text={{@badgeConfig.text}}
+          @skin={{@badgeConfig.skin}}
+          @size="sm"
+        />
+      {{/if}}
+      <div class="oss-stats-banner__content">
+        {{#if @loading}}
+          <OSS::Skeleton @height={{12}} @width={{150}} class="oss-stats-banner__skeleton--title" />
+        {{else}}
+          <div class="oss-stats-banner__title-row fx-row fx-gap-px-8 fx-xalign-center">
```
phndiaye: there is no gap-8 (spotted multiple btw) ^^

also, as there is already a class for it, might as well move the utility classes in there imo (applies to multiple cases below)

addon/components/o-s-s/stats/banner.hbs
phndiaye, line 1: Globally, i would add some data-control-name on the key blocks

addon/components/o-s-s/stats/banner.hbs
phndiaye: Personally, i would rather make a slightly more complex/future-proof title arg:

```
type StatTitle = {
  label: ReturnType<typeof IntlService['t']> | string;
  tooltip?: ReturnType<typeof IntlService['t']> | string;
}
```

This way we don't necessarily need a named-block but we still group this as belonging to the title

addon/components/o-s-s/stats/banner.ts
```diff
@@ -0,0 +1,21 @@
+import Component from '@glimmer/component';
+import type { OSSTagArgs } from '@upfluence/oss-components/components/o-s-s/tag';
+import type { OSSBadgeArgs } from '@upfluence/oss-components/components/o-s-s/badge';
+import type { OSSIconArgs } from '@upfluence/oss-components/components/o-s-s/icon';
+
+export type StatsBannerExtraInfoTagConfig = Pick<OSSTagArgs, 'label' | 'skin' | 'icon' | 'plain'>;
+export type StatsBannerBadgeConfig = Pick<OSSBadgeArgs, 'icon' | 'image' | 'text' | 'skin'>;
+export type StatsBannerExtraIconConfig = Pick<OSSIconArgs, 'icon'> & { color?: string };
+
+interface OSSStatsBannerSignature {
```
phndiaye, line 24: Is having everything optional here realistic? 🤔 I see at least title as being mandatory / its presence being validated

addon/components/o-s-s/stats/banner.hbs
```diff
@@ -0,0 +1,77 @@
...
+          />
+        {{/if}}
+        <OSS::Badge
+          @icon={{@badgeConfig.icon}}
+          @image={{@badgeConfig.image}}
+          @text={{@badgeConfig.text}}
+          @skin={{@badgeConfig.skin}}
+          @size="sm"
+        />
+      {{/if}}
```
phndiaye, line 20: Same as the possibly more complex `title`, I would say `badge` (and not `badgeConfig` as it's implicit imo):

```
type Badge = {
  badge: StatsBannerBadgeConfig;
  prefixIcon?: string // for the extra icon config as it's gated by badge presence
}
```

And i would say:

```hbs
- if badge named-block -> yield to it
- else if badge -> display it as you do now
- nothing otherwise
```

addon/components/o-s-s/stats/banner.hbs
```diff
@@ -0,0 +1,77 @@
...
+            </span>
+            <div class="oss-stats-banner__title-actions fx-row fx-gap-px-8 fx-xalign-center">
+              {{#if @titleInfoCircle}}
+                <span {{enable-tooltip title=@titleInfoCircle placement="top"}}>
+                  <OSS::Icon @icon="fa-info-circle" class="font-color-gray-400" />
+                </span>
+              {{/if}}
+              {{#if (has-block "dropdown")}}
+                {{yield to="dropdown"}}
+              {{/if}}
```
phndiaye: nit: is dropdown the best name here? technically you cannot enforce at all what the consumer will do w/ it ^^ maybe `title-suffix` named-block instead ? (if others have ideas 🙏🏾 )

app/styles/organisms/stats-banner.less
edouardmisset: 😍

addon/components/o-s-s/stats/banner.hbs
```diff
@@ -0,0 +1,77 @@
...
+          </div>
+        {{/if}}
+      </div>
+    </div>
+
+    {{#if (and (has-block "extra-badges") (not @loading))}}
+      <div class="oss-stats-banner__extra-badges fx-row fx-gap-px-3 fx-xalign-center">
+        {{yield to="extra-badges"}}
+      </div>
+    {{/if}}
```
phndiaye, line 57: can you link a design for this use case? i'm having trouble visualising what it maps to

addon/components/o-s-s/stats/banner.hbs
```diff
@@ -0,0 +1,77 @@
...
+            @plain={{@extraInfoTagConfig.plain}}
+          />
+        {{/if}}
+      {{/if}}
+    </div>
+    {{#if (and (has-block "cta") (not @loading))}}
+      <div class="oss-stats-banner__cta">
+        {{yield to="cta"}}
+      </div>
+    {{/if}}
```
phndiaye: wondering... should we keep this a named block or an `action: ButtonArgs & { fn: (...args: unknown[]) => unknown` (a bit like what we do for `OSS::Dialog` ? 🤔

addon/components/o-s-s/stats/banner.hbs
```diff
@@ -0,0 +1,77 @@
...
+        {{/if}}
+
+        {{#if @extraInfoTagConfig}}
+          <OSS::Tag
+            @label={{@extraInfoTagConfig.label}}
+            @skin={{@extraInfoTagConfig.skin}}
+            @icon={{@extraInfoTagConfig.icon}}
+            @plain={{@extraInfoTagConfig.plain}}
+          />
+        {{/if}}
```
phndiaye: in the same logic as above, what about:

```
type StatValue = {
  label: ReturnType<typeof IntlService['t']>, // or value
  suffix?: ReturnType<typeof IntlService['t']>,
  tags?: TagArgs[] // worst case there is always only one but we support multiple (not mandatory though, could be just `tag`)
}
```

tests/dummy/app/templates/index.hbs
phndiaye, line 1: This would fit better in the `visual` or `data` route i think :)

phndiaye (commented):
Global: kinda late on the wording but what about `OSS::Stats::Card` rather than `Banner` ? 🤔  (unless it maps to the Figma wordings of course :()

app/styles/organisms/stats-banner.less
edouardmisset: ... But you're missing `container-type: inline-size` in the parent in order to enable this ;)

edouardmisset: https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries

addon/components/o-s-s/stats/banner.hbs
```diff
@@ -0,0 +1,77 @@
...
+      {{/if}}
+    </div>
+    {{#if (and (has-block "cta") (not @loading))}}
+      <div class="oss-stats-banner__cta">
+        {{yield to="cta"}}
+      </div>
+    {{/if}}
+  </div>
```
edouardmisset, line 90: question: where is the "normal" / "default" `{{ yield }}` to pass the leftovers? I might very well be missing something here.

addon/components/o-s-s/stats/banner.ts
```diff
@@ -0,0 +1,21 @@
+import Component from '@glimmer/component';
+import type { OSSTagArgs } from '@upfluence/oss-components/components/o-s-s/tag';
+import type { OSSBadgeArgs } from '@upfluence/oss-components/components/o-s-s/badge';
+import type { OSSIconArgs } from '@upfluence/oss-components/components/o-s-s/icon';
+
+export type StatsBannerExtraInfoTagConfig = Pick<OSSTagArgs, 'label' | 'skin' | 'icon' | 'plain'>;
```
Elodie-DeMatteis-Upf: nit: Why did you suffix every arg with "config"? I feel like it may be too much, especially as it's repeated in every arg

addon/components/o-s-s/stats/banner.hbs
OwenCoogan: Updated

addon/components/o-s-s/stats/banner.hbs
OwenCoogan: Updated

tests/integration/components/o-s-s/stats/banner-test.ts
OwenCoogan: Added test

tests/integration/components/o-s-s/stats/banner-test.ts
OwenCoogan: added tests

addon/components/o-s-s/stats/banner.ts
OwenCoogan: Updated

addon/components/o-s-s/stats/banner.hbs
OwenCoogan: Updated

addon/components/o-s-s/stats/banner.hbs
OwenCoogan: Updated

addon/components/o-s-s/stats/banner.hbs
OwenCoogan: https://www.figma.com/design/6I3IoYJkvAOOs7sPH9L8gt/Stats-card-redesign?node-id=19606-1698&t=muqAvXMEkjKivDa1-11

integrations badges, already tried out in the implem pr from my slack message 👌

addon/components/o-s-s/stats/banner.hbs
OwenCoogan: On this one we need to leave it open as a named block, component specs specify that there is a possiblity for secondary actions

tests/dummy/app/templates/index.hbs
OwenCoogan: moved it to data

addon/components/o-s-s/stats/banner.hbs
OwenCoogan: We don't really need it here, not unless we want to pass other components on the fly but that's a whole other implem IMO ^^

addon/components/o-s-s/stats/banner.ts
OwenCoogan: Updated

Miexil (approved):

addon/components/o-s-s/stats/banner.hbs
edouardmisset: maybe add it to the storybook also ?

edouardmisset (approved):

Elodie-DeMatteis-Upf (approved):

